PR #7, opened 2019-08-09
Open dnsmasq to containers

fooman:
Prior to this change I had problems with DNS resolution from within my custom container in the warden network. With this change I can see the outbound queries (for example bitbucket.org) in the output of `docker logs dnsmasq`.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -36,7 +36,7 @@ services:
     container_name: dnsmasq
     image: jpillora/dnsmasq
     ports:
-      - "127.0.0.1:53:53/udp"
+      - "53:53/udp"
     volumes:
       - ~/.warden/etc/dnsmasq.conf:/etc/dnsmasq.conf
     labels:


davidalger:
@fooman This is an interesting use-case and sounds like a rather good idea, although it actually would  revert a change made in commit 8e16d095 which fixed a problem running Warden on Ubuntu (which will not allow the port 53 bind to 0.0.0.0, but does to 127.0.0.1).

As of the 0.1.6 release, which I literally just pushed up this morning with changes made earlier this week, the `php-fpm` and `php-debug` containers are now members of the `warden` docker network. That was done to support SoapClient requests sending things to Traefik, but incidentally might allow for also using `dnsmasq` (by it's container IP address) for the resolver within the container. If you're setting up a custom `/etc/resolve.conf` file within the container, could also put the DNS on a secondary port and add a `port 6053` to that file.

How are you publishing the DNS resolver into the container? Via a `/etc/resolve.conf` built into the image or via `dns: [ <address> ]` in the docker compose file?

fooman:
Ah tricky.

I am currently trying to use my custom images with Warden (replacing my existing dnsmasq + jwilder/nginx-proxy solution).

In the overall set up I believe the trickiest part is to run functional tests from within the same environment. The selenium node will need to be able to resolve for example magento2.warden.test. This info is only in the dnsmasq container on the host.

So far my approach has been to apply an ip alias on the host machine and then supply this IP address as the dns to the container:
```
docker run --dns  10.254.254.254 --network warden --rm byrnedo/alpine-curl  https://www.google.com
```
>curl: (6) Could not resolve host: www.google.com


Your suggestion to use the IP address of the dnsmasq container instead works
```
docker run --dns 172.22.0.5 --network warden --rm byrnedo/alpine-curl  https://www.google.com
```

or without relying on a fixed IP
```
 docker run --dns `docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' dnsmasq` --network warden --rm byrnedo/alpine-curl  https://www.google.com
```

I'll close this one as it now works for me with the above suggestion.

Short of using a fixed IP for the dnsmasq container I haven't seen any other docker/docker-compose options so far. See for example this feature request https://github.com/docker/cli/issues/385.

If you are open to it I could try do add a
dns: DNSMASQ IP
entry similar to how you did it with extra_hosts in the env command.

PR #18, opened 2019-10-17
Add note to readme to assign docker >4GB RAM

Vinai:
Reference thread https://twitter.com/blackbooker/status/1184638003953754113


davidalger:
Thanks for adding this so it won’t be forgotten!

PR #19, opened 2019-10-17
Add screenshots for PHPStorm DB config

Vinai:
Maybe it would be a good idea to put the new images into the same CDN that is used for the other images.


Vinai:
In fact, I'll update the PR to point to the github CDN :)

davidalger:
Thanks! The Github CDN is perfect (I may move the others from where they’re at currently, which is a personal S3 bucket I use for for sharing screenshots)

Issue #33, opened 2019-10-29
Automatically trigger Xdebug when running PHP on CLI inside php-debug container


davidalger:
The php CLI inside the `php-debug` container should trigger Xdebug automatically. Currently the only way to debug CLI commands (as opposed to HTTP requests which are routed to Xdebug automatically already) is to specify a couple settings to php:

```
warden env exec php-debug bash

php -d xdebug.remote_autostart=on -d xdebug.remote_host=host.docker.internal ...
```

At minimum remote_autostart should be set on CLI by default, and ideally the remote_host as well. That one will require a little work to get right on both macOS and Linux.

davidalger:
> 10/18/19, 10:10 AM
⁦‪@blackbooker‬⁩ ⁦‪@VinaiKopp‬⁩ ⁦‪@happyryanf‬⁩ ⁦‪@_Talesh‬⁩ I pass the values to the container environment variables via docker compose and -e.
e.g. docker-compose run -e PHP_IDE_CONFIG="serverName=$server_name" -e XDEBUG_CONFIG="remote_host=172.17.0.1" magento-cli "$@" now i just use this command instead of running with php flags
ref: https://twitter.com/petey_tweety/status/1185211571493994496

blackbooker:
Jonathan Vale
General Partner
602-769-1093
Vale Investment Group, LLC.

molotovbliss:
https://gist.github.com/molotovbliss/3eb18a2b0366d9b9e15647322641816d just to note for a possible way to get the container id along with passing the usual xdebug php params & script file to debug.

wojtekn:
@davidalger this sounds as a great idea.

I'm not sure if I'm missing something, but wouldn't it be just about adding those two config directives in /etc/php.d/15-xdebug.ini in this container?

davidalger:
Per a conversation I had with @wcrb15 the gateway IP of the `warden` network works as the remote_host on Linux systems. So in order to resolve this, it should be possible to look up and pass this gateway IP address as an extra_hosts parameter on Linux systems allowing docker.host.internal to resolve there as well as on macOS within the php-debug container. For the sake of not disrupting any existing functionality, the use of connect back for browser requests should remain untouched.

The gateway IP address can be looked up via the following similar to [what is done here](docker container inspect traefik --format '{{.NetworkSettings.Networks.warden.Gateway}}'):
```
docker container inspect traefik --format '{{.NetworkSettings.Networks.warden.Gateway}}'
```

[...]

PR #35, opened 2019-11-13
Issue #33: Add ability to call xDebug via warden for CLI scripts

molotovbliss:
Feature request enhancement #33


davidalger:
@molotovbliss Thanks for this! I will try and take a good look at it some time next week. This highlights the need I think (referencing the gist you linked on #33) for adding the ability to setup custom commands on an installation of Warden as well.

molotovbliss:
> @molotovbliss Thanks for this! I will try and take a good look at it some time next week. This highlights the need I think (referencing the gist you linked on #33) for adding the ability to setup custom commands on an installation of Warden as well.

Np, wasn't sure if there was a better way to call `docker` exec via warden as it seems most go towards `docker-compose`. BTW, warden makes Magento dev a breeze. Thanks for the great tool.

mike-keilty:
@molotovbliss I tested these changes out locally and they work great!  Thank you for this.

molotovbliss:
@mike-keilty If it helps just one person that's awesome in my book. BTW I believe the next warden release has a much more elegant approach than this. (See the referenced issue associated with this pr). Cheers! 🍺

davidalger:
This PR has been rebased on latest develop and updated to conform to the following (slightly different) take on the original proposed solution:

1. Introduces a new command `warden debug`; when this is run the user is launched into a shell similarly to when `warden shell` is run, but in this case the shell is opened by default within the php-debug container.

2. When user is launched into a debug container, the env variable XDEBUG_REMOTE_HOST is passed in and conditionally set to either `host.docker.internal` on macOS and to the gateway IP of the warden network on Linux systems.

3. The base images have been updated to add the following to `/etc/php-cli.ini` bringing it all together without any potential for disrupting already present and reliable in-browser debugging:

```
[php]
xdebug.remote_autostart=on
xdebug.remote_host=${XDEBUG_REMOTE_HOST}
```

Big shout out to @wcrb15 for helping ascertain what the remote host should be configured to on Linux systems and to @molotovbliss for getting the ball rolling with the idea of a special debug command.

PR #40, opened 2019-11-20
Selenium Standalone instance as an option

lbajsarowicz:
Add `WARDEN_SELENIUM` option to install Selenium Standalone which is compatible with Magento 2 MFTF


davidalger:
@lbajsarowicz This has been published in Warden 0.1.11. Thank you for the contribution!

PR #46, opened 2019-12-06
Update README.md

ryantfowler:
Included reference to one solution which includes a M2 application-level solution for decyphering inbound hostnames.


davidalger:
Thanks @ryantfowler !

Issue #53, opened 2019-12-24
Check for existence of /etc/dhcp/dhclient.conf during install script


wcrb15:
I was trying to get Warden to run inside of Manjaro/Arch Linux and by default the /etc/dhcp/dhclient.conf file does not exist. The install command should check for the existence of this file and if it doesn't exist try to create it. In this particular use case I had to also configure Network Manager to use dhclient as its DHCP backend, but that is expected since Arch isn't officially supported. I will probably write up a PR for this soon but wanted to officially open an issue about it.

PR #54, opened 2019-12-24
Added check to ensure conf file exists before using it

wcrb15:
This is a quick change to make sure that the conf file exists before using it. That should fix up the issue I opened about installing on Arch

### DIFF
--- a/commands/install.cmd
+++ b/commands/install.cmd
@@ -76,8 +76,12 @@ if [[ "$OSTYPE" == "linux-gnu" ]] && [[ ! -f "${WARDEN_HOME_DIR}/nodnsconfig" ]]
     echo "==> Configuring resolver for .test domains (requires sudo privileges)"
     if ! sudo grep '^prepend domain-name-servers 127.0.0.1;$' /etc/dhcp/dhclient.conf >/dev/null 2>&1; then
       echo "  + Configuring dhclient to prepend dns with 127.0.0.1 resolver (requires sudo privileges)"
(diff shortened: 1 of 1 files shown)


wcrb15:
#53 is the issue I opened for this

commands/install.cmd
```diff
@@ -74,6 +74,8 @@ if [[ "$OSTYPE" == "linux-gnu" ]] && [[ ! -f "${WARDEN_HOME_DIR}/nodnsconfig" ]]
     echo "==> Configuring resolver for .test domains (requires sudo privileges)"
     if ! sudo grep '^prepend domain-name-servers 127.0.0.1;$' /etc/dhcp/dhclient.conf >/dev/null 2>&1; then
       echo "  + Configuring dhclient to prepend dns with 127.0.0.1 resolver (requires sudo privileges)"
+      ## Ensure that dhclient.conf exists before using it
+      mkdir -p "/etc/dhcp" && touch "/etc/dhcp/dhclient.conf"
```
davidalger: Would it make more sense perhaps to add a condition for file presence to the if conditional which checks for the prepend directive? I.e. don't do anything if the file does not exist. Or is the file not there by default on Arch linux, and once it's there it'll be utilized, one just has to create it first?

wcrb15: Yeah by default that file doesn't exist at all in Arch and will be used once the file is created and Network Manager is configured to use dhclient as its DHCP backend.

PR #64, opened 2020-01-07
Created modular warden

rafashkembi:
Warden is a very powerful tool and i really like how it works.
A lot of agencies want to have custom environments and commands that can be used with warden.
Since this was not possible I have modified the files and created two new folders so the developers can add new environments and commands.


fooman:
Custom environments are already somewhat supported. In your project folder create `.warden/warden-env.yml` and in there you can supply your own images, etc.

jamescowie:
I like the idea. As @Fooman says custom environments are for most use cases handled via the .warden/warden-env.yml where additional docker services can be created and linked. 

What I like about this PR is that for our use case internally we need to add custom warden commands to fit the workflow we are creating. Where this comes from and its easier to share the big picture is that we are using warden now as our internal workflow tool, So instead of having many bash scripts that we all need to install that deals with importing DB's etc we want to extend warden to allow us to add these in. Some we will share back with the world others are specific to shero that would make no sense. 

Thats why I like that creating the ~/.warden/commands/new-command.sh and letting warden load additional commands will be really cool.

fooman:
Just wondering if it might be an option to host the new commands in a custom docker image which then gets spun up alongside the other warden images with access to the other container's volumes.

davidalger:
@rafashkembi Very much appreciate the POC demonstrating this concept for everyone. I'm sorry I did not get to addressing this sooner. It made it tricky since when the PR was first opened, I was thinking through the refactoring that would eventually become 0.5.0 a few months later (as I had the time to make it happen)

As far as custom environment types go, definitely take a look at the changes on 0.5.0 and it's ability to add services like redis for example to a 'local' env type combined with .warden/warden-env.yml for per-project definitions and provide feedback if that doesn't cut it.

For implementing the custom commands, check out PR #172 where I've re-implemented this concept to support both custom commands in ~/.warden/commands AND a project's .warden/commands directory. Link to an example implementation of a custom command can be found there as well.

Issue #70, opened 2020-01-11
Support for custom backend URL


lbajsarowicz:
Warden should support custom Admin domain, for example:
`backend.enterprise.test`

davidalger:
Closing issue for now due to inactivity. Also worth mentioning, since Warden 0.2.0 a wildcard rule has been configured on Traefik to route *.example.com to the project environment with example.com also routing (apex domain) to the project environment as of Warden 0.4.4.

Issue #67, opened 2020-01-08
Change selenium environment include to use hub/node configuration


davidalger:
As configured currently selenium is using an all-in-one container, this would preclude multi-browser testing. Would like to see something like the following (i.e. hub and a chrome node) supported OOTB here (allowing environments to add additional nodes as needed).

```
  selenium-hub:
    container_name: ${WARDEN_ENV_NAME}_selenium-hub
    image: selenium/hub:3.141.59-titanium
    ports:
      - "4444:4444"
  selenium-chrome:
    container_name: ${WARDEN_ENV_NAME}_chrome
    image: selenium/node-chrome:3.141.59-titanium
    volumes:
      - /dev/shm:/dev/shm
    depends_on:
      - selenium-hub
    environment:
      - HUB_HOST=selenium-hub
      - HUB_PORT=4444
      - START_XVFB=false

```
Note this will require reform before dropping into place as it came from a sidecar stack.

cc @lbajsarowicz would be wonderful if you'd like to tackle this since you've got a working MFTM setup you could validate the changes with, otherwise I'll probably tackle it only if I have the time leading up to Warden 0.2.0 release.

lbajsarowicz:
Good point, but we need to take into consideration the fact that Codeception is not supporting Selenium that is more recent than `3.8.1`. What is more - I'd like to give choice to use DEBUG version (that can be connected with VNC), not only `-titanum` one without possibility to check what's wrong.

lbajsarowicz:
Another point is... Allure reporting. We should be able to provide interface - for example allure.{domain}.test

PR #71, opened 2020-01-11
Support for custom Admin subdomain

lbajsarowicz:
Implementation for #70, possibility to define custom Admin subdomain.

Pull Request #67 has to be merged first, as this PR also covers adding custom hostname to Selenium.


lbajsarowicz:
@davidalger I need your help with this section:
https://github.com/davidalger/warden/blob/aeed219905fb59b5b67438791b3d71c47d8ec009/environments/magento2.admin-domain.base.yml#L3-L5

I don't like my approach, but I couldn't do it differently (no skills, actually).
I would like to **extend** by adding extra line instead of **overriding** by rewriting existing rule :-(

davidalger:
This should actually be addressed already, although a documentation update lining out how to configure custom admin domain for Magento may be in order now.

With the latest code already on `develop` for the upcoming Warden 0.2.0 release, Traefik is being configured with a wildcard match now, allowing any sub-domain of the project domain to route to Varnish/Nginx by default without any changes to the project. So the only thing required to use a custom admin domain would be configuring it in Magento configuration and/or env.php file. 

It's fairly impossible to have `extra_hosts` include a wildcard/unknown admin domain given anything there is simply written into `/etc/hosts`. I'm thinking the only record needed there however would be the primary URL right? It's there so Magento can properly request it's own WSDL file when SOAP API is used. Does MFTF also require ability to hit the admin URL?

lbajsarowicz:
@davidalger MFTF hits the Admin URL.

PR #80, opened 2020-01-29
Add Cron to Magento 2

navarr:


environments/magento2.darwin.yml
```diff
@@ -21,5 +21,16 @@ services:
       - .${WARDEN_WEB_ROOT:-}/pub/media:/var/www/html/pub/media:delegated
       - appdata:/var/www/html
 
+  crontab:
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
```
davidalger: This should likely be marked as read-only `:ro`. Also with the docker.sock being mounted, this container will need to be marked as privileged or it will not work on Linux installations with SELinux or AppArmor enabled.

environments/magento2.base.yml
```diff
@@ -136,6 +136,39 @@ services:
       - warden
       - default
 
+  crontab:
+    hostname: "${WARDEN_ENV_NAME}-crontab"
+    image: navarr/container-crontab:v0.1.0
```
davidalger: Where can I find the Dockerfile for this image?

environments/magento2.base.yml
```diff
@@ -136,6 +136,39 @@ services:
       - warden
       - default
 
+  crontab:
```
davidalger: Would it work to add this as a Global Service rather than have a crontab per environment? If I'm understanding things correctly, this is looking at labels on running containers and essentially `exec`'ing into them to run what's been defined in the labels.

environments/magento2.base.yml
navarr: https://github.com/navarr/container-crontab

environments/magento2.base.yml
navarr: I considered that but not without heavy modifications to the code (which I'm not comfortable doing).

The reason being is that if it was a global service, it would run any cron that had been started at one point after the image.  A `warden env down` would not work, as the global crontab would restart the cronjob of that env.

environments/magento2.darwin.yml
navarr: My assumption is that it uses `/var/run/docker.sock` not only to listen but also to trigger exec of the images - but I don't understand enough of the docker API to know that for sure.

davidalger:
One other question: What will happen when these containers are started, but Magento code does not yet exists in the containers (as would be the case while progressing through an initial setup)?

navarr:
> One other question: What will happen when these containers are started, but Magento code does not yet exists in the containers (as would be the case while progressing through an initial setup)?

Fantastic question - I did not think to check that.  I imagine the cronjob container would start and fail, regularly, every minute, until it did go up.  I don't really foresee that being a problem, but if it is I could probably add a check for bin/magento to exist before running.  It would still fail if the site was not yet installed, however, and I'm not sure how to check that.

environments/magento2.base.yml
davidalger: That makes sense. I think `down` would be ok (it destroys the container) but `stop` would not since they are still there. I spent a little time reading through the code for it myself late on Friday and thought it looked like it was listening to stop/die events to de-register, but when I ran a little test, it panned out as you described, so I'm not 100% what the event logic was for (possibly something more directly tied to Rancher behaviour). Since it's either starting/restarting/stopping to execute a job, it makes sense it'd be difficult to alter the code in this way.

Found this little gem which is similar: https://github.com/mcuadros/ofelia but it doesn't currently watch docker events (only reads what's there when it starts up) so it wouldn't work in this way either (i.e. one orchestration container looking at labels across all containers).

Going to try and run another test this week and see what happens if I have Warden remove the cron job container in an environment entirely when running `stop` and see what happens. That might work to allow using this, but with a single orchestration container.

navarr: It sounds like my being nerd sniped has nerd sniped you

davidalger:
Wanted to leave a note here saying I really truly appreciate the effort that went into putting this PR together. At the end of the day, I think both of us learned a lot as a result. However after much deliberation I'm going to close this out in favor of taking a simpler and more portable (i.e. between applications) route of installing `crontabs` in the FPM images and running `crond` in the background via the entrypoint script.

There are a few reasons leading me to go down this path:

* Portability, as expected crontab definitions will not be needed on a per-application environment basis.
* Projects with custom cron jobs (as is often the case with Symfony apps) would require custom docker-compose configuration, and that makes it more difficult for developers to use them when needed.
* Magento has a `bin/magento cron:install -d` command which adds the following to the crontab and I'd like to eliminate "why doesn't this work" questions:

```
www-data@rock-php-fpm:/var/www/html$ crontab -l

#~ MAGENTO START 69dd2b02e1f3a65918182048ea4e29979a849d8942e8f53ed20a4bf10e529b36
* * * * * /usr/bin/php /var/www/html/bin/magento cron:run 2>&1 | grep -v "Ran jobs by schedule" >> /var/www/html/var/log/magento.cron.log
#~ MAGENTO END 69dd2b02e1f3a65918182048ea4e29979a849d8942e8f53ed20a4bf10e529b36

```

* If need be this custom crontab can be persisted (across container re-creations) by mounting a file into the container at `/var/spool/cron/www-data` by adding custom configuration to `.warden/warden-env.yml` such as:

```
version: "3.5"
services:
  php-fpm:
    volumes:
      - .warden/crontab:/var/spool/cron/www-data
```

* Lastly I couldn't get past the belief that hard-coding a specific cron routine in a base environment runtime configuration was the right move given the disparate nature of crons, and also the potential danger of having crons automatically setup and running.
* Given this is a development environment, not a distributed production container runtime, I think it's ok to have a secondary process running in the FPM container for crons given the complexity reduction it offers here.

The FPM images have been updated in commit eab353f.

navarr:
No complaints. 👍

PR #94, opened 2020-02-13
Fixes constraint verification of mutagen version

blakesaunders:
Fixed validation condition where >=0.11.0 <0.11.2 would fail, although warden is requesting a mutagen version >=0.10.3.


davidalger:
@blakesaunders Huge thank you for this! Ran a few checks locally and it works correctly in every case I tried. Going to cut release 0.2.3 with this fix shortly.

davidalger:
Published in Warden 0.2.3 release.

PR #103, opened 2020-02-20
Docker memory preferences path updated

snowcore:
The path was changed to _Preferences -> **Resources ->** Advanced -> Memory_


davidalger:
Thank you @snowcore!

PR #107, opened 2020-02-25
FIX: Timeout running MFTF tests

lbajsarowicz:
I've ran into 2 kinds of timeouts:
1. From Selenium Grid
2. From PHP endpoint of MFTF

image

This PR fixes both types of errors


davidalger:
@lbajsarowicz Can you add a line for this to the changelog? Would be happy to get this merged. The image change will build as soon as the PR merges and the other alteration will make it into next release or be available using `--HEAD` on the brew installation in the meantime.

PR #116, opened 2020-03-17
Feature: `warden vnc` to connect with Selenium Debug VNC

lbajsarowicz:
I've introduced a new Warden command that enables developer to connect VNC using `remmina`:

```bash
$ warden vnc
> We are running connection for you. Password is "secret".
```

Additionally, documentation extended with such information.

PS: @davidalger I added Color Reset after each Error message.


lbajsarowicz:
I've added multi-node and different types support:
`warden vnc 5` opens 5th instance of the Selenium node.
`warden vnc 1 selenium-firefox` opens 1st instance of Selenium Firefox node

davidalger:
I've integrated the latest origin/develop into this PR and updated the messaging to use the new messaging functions, as well as tweaking a few details so they're compatible with latest 0.6.0 develop branch.

The following is what I'd like to see happen for this to be merged:

Since Warden is not Linux specific I would prefer to not adopt commands as 1st class citizens unless they are interoperable between both OS's. With Remina not supporting Mac OS, this command needs to do something to support Mac OS apart from using Remina. Even a graceful fallback would be acceptable, as long as it's mentioned in the documentation page. For example, the ssh tunnel currently mentioned in the docs page for use on OS X would suffice if the command printed out connection details for the user, and then started the ssh tunnel (preferably in the foreground so the user can Control + C to kill it):

```
ssh -N -L localhost:5901:magento2_selenium-chrome_1:5900 tunnel.warden.test
```

On linux environments, the command could print a warning (using the `warning` function) when Remina is not present, and fall back to the same ssh tunnel approach allowing the nicety of using Remina on Linux to be supported, but also having an OS agnostic approach which would work as a fallback.

I'd also like to see a `commands/vnc.help` added with basic usage information (see the others present, doesn't have to be fancy).
*

lbajsarowicz:
Version when no Remmina installed
image

PR #126, opened 2020-03-23
Add instructions for connecting to VNC server using Mac OS X

navarr:


davidalger:
Before I merge this, I want to check and see if a TCP proxy could be configured using labels on the Selenium container allowing `vnc://selenium.exampleproject.test` to be used on both linux and macOS without needing to have Reminas tunnel via SSH or require the SSH tunnel be created on macOS. This thought is what has held me back on merging #116 as well, which wouldn't work on macOS since Reminas does not exist (to my knowledge) for macOS at this time.

navarr:
In order to use the hostnames, each instance would need its own proxy (since afaik VNC doesn't have any concept of SNI)

I tried to get some sort of TCP routing working through Traefik but so far have been entirely unsuccessful at getting Traefik to identify my TCP router labels or even acknowledge them in its logs.

navarr:
I would think, in the interim, updating the docs for the current usage is still a good idea ;)

davidalger:
True…I won't let it hold this up, but the PR introducing a Linux only cmd to Warden will need to wait till I can look into that. Good notes. Thanks!

PR #141, opened 2020-04-14
Use Mutagen pause/resume on stopped environments

fooman:
Mutagen is able to pause and resume sync sessions. I added pause/resume/monitor to the `warden sync` command. Additionally I changed the automatic Mutagen behaviour for the warden environment commands.

Now the following should happen:

- `warden env up -d` when run for the first time should trigger `warden sync start` as before
- `warden env down` should trigger `warden sync stop` as before
- `warden env stop` should now trigger `warden sync pause` (you can check the status should say paused under `warden sync list`
- `warden env up -d` when run for a previously stopped environment should trigger `warden sync resume` provided the underlying php-fpm container still has the same container id (you can check `warden sync list` should have the same synchronisation identifier) and we have a paused session

### DIFF
--- a/commands/env.cmd
+++ b/commands/env.cmd
@@ -83,10 +83,28 @@ export TRAEFIK_ADDRESS="$(docker container inspect traefik \
     ' 2>/dev/null || true
 )"
 
+## pause mutagen sync if needed
+if [[ "${WARDEN_PARAMS[0]}" == "stop" ]] \
+    && [[ $OSTYPE =~ ^darwin ]] && [[ -f "${WARDEN_DIR}/environments/${WARDEN_ENV_TYPE}/${WARDEN_ENV_TYPE}.mutagen.yml" ]]
+then
+    warden sync pause
+fi
+
 ## anything not caught above is simply passed through to docker-compose to orchestrate
 docker-compose \
     --project-directory "${WARDEN_ENV_PATH}" -p "${WARDEN_ENV_NAME}" \
     "${DOCKER_COMPOSE_ARGS[@]}" "${WARDEN_PARAMS[@]}" "$@"
+
+## resume mutagen sync if available and php-fpm container id hasn't changed
+if ([[ "${WARDEN_PARAMS[0]}" == "up" ]] || [[ "${WARDEN_PARAMS[0]}" == "start" ]]) \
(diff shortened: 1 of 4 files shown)


commands/env.cmd
```diff
@@ -83,11 +83,29 @@ export TRAEFIK_ADDRESS="$(docker container inspect traefik \
...
     --project-directory "${WARDEN_ENV_PATH}" -p "${WARDEN_ENV_NAME}" \
     "${DOCKER_COMPOSE_ARGS[@]}" "${WARDEN_PARAMS[@]}" "$@"
 
+## resume mutagen sync if available and php-fpm container id hasn't changed
+if ([[ "${WARDEN_PARAMS[0]}" == "up" ]] || [[ "${WARDEN_PARAMS[0]}" == "start" ]]) \
+    && [[ $OSTYPE =~ ^darwin ]] && [[ -f "${WARDEN_DIR}/environments/${WARDEN_ENV_TYPE}/${WARDEN_ENV_TYPE}.mutagen.yml" ]] \
+    && [[ $(warden sync list | grep -i 'Status: \[Paused\]' | wc -l | awk '{print $1}') == "1" ]] \
+    && [[ $(warden env ps -q php-fpm) ]] \
+    && [[ $(docker container inspect $(warden env ps -q php-fpm) --format '{{ .State.Status }}') = "running" ]] \
+    && [[ $(warden env ps -q php-fpm) = $(warden sync list | grep -i 'URL: docker' | awk -F'/' '{print $3}') ]]
```
fooman, line 104: One thing I am unsure about is if we need to use `=` or `==`. The existing implementation seems to use both and my bash knowledge is limited as to what to prefer here.

davidalger:
This looks super neat. I'll run it through a few cycles on my macOS setup and see how it checks out.

commands/env.cmd
davidalger: The short version is `=` is the original equality operator, and at some point bash (and other shells) added `==` as an alias. They are both lexical string comparison operators when enclosed in `[ ]` but this changes when you enclose them in `(( ))` (arithmetic operator) where the two are no longer synonymous.

fooman:
Good stuff - sorry the steps to confirm where off (`warden down` instead of `warden env down` and `warden stop` instead of `warden env stop`) but you probably already caught that.

davidalger:
Amazing work. Tested out Ok, so I got it merged and included in the 0.4.2 release which I've just published. Cheers!

PR #146, opened 2020-05-06
Add support for Symfony 4+

lbajsarowicz:
This PR adds the skeleton to run Symfony 4+ on Warden OOTB.
Due to the fact that Symfony uses `.env` file, too - I used `WEBROOT` with value of `web/` by default


davidalger:
@lbajsarowicz Thanks for this! I'll be pushing a similarly named branch to this repo with this work merged into the `feature/environment-template-restructure` branch where I'm working on the implementation outlined here (which is for the 0.5.0 release): https://github.com/davidalger/warden/pull/105#issuecomment-628362330

Few points of feedback:

* We can't use an appdata volume on the macOS side unless there is a Mutagen configuration for Symfony in place. I'll be stripping that out to use a delgated mount on macOS similar to Laravel and Magento 1 types (these may all eventually get Mutagen configurations if there is reason and request for it)
* I'll be removing the custom set of `WARDEN_WEB_ROOT`; understand the reasoning behind that note, but at this point I'm not ready to have new environments use a custom webroot location (docs would need to be updated, and I'd want it to be uniform across all env types should this take place, reason here being the intent is to have it be really easy to add a warden configuration to an existing project) - I do understand this currently poses a challenge with committing the .env for Warden but want to take this one step at a time.

Once I finish the work on https://github.com/davidalger/warden/tree/feature/environment-template-restructure I'll merge https://github.com/davidalger/warden/tree/feature/symfony-support into develop as well, and when I do that, this PR will close. I can't update the branch on your repo due to permissions, so if you want to test the integrated work ahead of time, feel free to poke at that branch prior to it being merged (it probably won't be until next week after I'm back from vacation)

davidalger:
To compare `feature/symfony-support` against new base `feature/environment-template-restructure` branch go here: https://github.com/davidalger/warden/compare/feature/environment-template-restructure..feature/symfony-support

davidalger:
Added note regarding new Symfony env type on this docs page: https://docs.warden.dev/environments/types.html

Thanks for the contribution @lbajsarowicz

PR #154, opened 2020-05-20
Give composer all the memory

navarr:
Increasingly commonly Composer is requiring more than the PHP memory_limit, this change prevents that issue by allowing composer to exceed the memory limit


davidalger:
Thanks @navarr!

erikhansen:
Great PR! I frequently have had to run:

```
php -d memory_limit=-1 /usr/bin/composer update
```

PR #170, opened 2020-06-22
Prompt unspecified env init inputs

Lunaetic:
PR replaces a TODO: with bash scripting that prompts the user for a warden environment name and a warden environment type if they were not provided when `env-init` was called.

**Testing** 

- Pull PR
- Create a new, empty directory, and from that directory, run `warden env-init test magento2`; assert that the `.env` file is created as normal
- Delete the .env file created above
- Run `warden env-init`; when prompted for an environment name, hit enter - you should be prompted until you provide a name. Provide one.
- You should be prompted for an environment type, hit enter - you should be prompted until you provide a valid environment type. Provide one.
- Assert that the generated .env file contains the correct environment name and environment type

### DIFF
--- a/commands/env-init.cmd
+++ b/commands/env-init.cmd
@@ -19,15 +19,23 @@ fi
 WARDEN_ENV_NAME="${WARDEN_PARAMS[0]:-}"
 
 # If warden environment name was not provided, prompt user for it
-if [ -z "$WARDEN_ENV_NAME"]; then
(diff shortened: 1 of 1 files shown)


davidalger:
Thank you @Lunaetic! I tweaked this slightly in 7af516aaf. It will be included in the soon-to-be-published 0.6.0 release. 🎉

PR #185, opened 2020-07-09
Fixes dnsmasq on Fedora systems

superterran:
points networkmanager to dhclient service to enable dnsmasq support on fedora


davidalger:
Thanks for the PR @superterran! I'm strongly considering stripping out the attempted auto-configuration of DNS resolver on Linux both to not muck with users configuration (which can be very a personal thing on Linux systems) as well as to resolve the warnings the systemd checks end up issuing when you try and run Warden either within a docker container (for DIND setups) and on WSL2.

The replacement would be something along the lines of this documentation from another project: https://devilbox.readthedocs.io/en/latest/intermediate/setup-auto-dns.html

superterran:
I think that's a good call, I've tested this config for about a week in Fedora and while it works for warden, it removes the ability to discover linux hostnames on the local network. This seems to be true for Ubuntu as well.

PR #186, opened 2020-07-09
Fixes TLS in Fedora

superterran:
moves cert to correct location for Fedora 32 and changes update command to what is suggested in man page

### DIFF
--- a/commands/install.cmd
+++ b/commands/install.cmd
@@ -32,9 +32,8 @@ if [[ "$OSTYPE" =~ ^linux ]] \
   ## Fedora/CentOS
 then
   echo "==> Trusting root certificate (requires sudo privileges)"  
-  sudo cp "${WARDEN_SSL_DIR}/rootca/certs/ca.cert.pem" /etc/pki/ca-trust/source/anchors/warden-proxy-local-ca.cert.pem
-  sudo update-ca-trust
-  sudo update-ca-trust enable
+  sudo cp "${WARDEN_SSL_DIR}/rootca/certs/ca.cert.pem" /usr/share/pki/ca-trust-source/anchors/warden-proxy-local-ca.cert.pem
+  sudo update-ca-trust extract
 elif [[ "$OSTYPE" =~ ^linux ]] \
   && [[ -d /usr/local/share/ca-certificates ]] \
   && [[ ! -f /usr/local/share/ca-certificates/warden-proxy-local-ca.crt ]] \


davidalger:
@superterran I'm not sure from the changes or the description what this is intending to fix. On any recent Fedora installation, the before/after result here is going to be 100% identical from a trust standpoint.

Per [Fedora docs](https://docs.fedoraproject.org/en-US/quick-docs/using-shared-system-certificates/) and the man page (for update-ca-trust) custom CA roots should either be added to `/etc/pki/ca-trust/source/anchors` OR `/usr/share/pki/ca-trust-source/anchors` Apart from precedence, the two locations operate identically. This is confirmed in RHEL documentation as well.

The reason for the use of `update-ca-trust` followed by `update-ca-trust enable` is due to older EL 6 systems. Probably don't have any of those using Warden, but on EL 6 the `update-ca-trust` was a bit more complicated than it is on recent EL and Fedora releases. On EL 6 the setup could actually require the `enable` call, and the script had a bunch of different verbs with no-verb being different from explicit `extract` merely in verbosity. Although the man page disagrees, the Fedora docs page (linked above) states one should simply call `update-ca-trust` (and I believe this is due to the abandonment of the use of params in the update-ca-trust command)

On recent EL 8 and Fedora systems, the `update-ca-trust` script is pretty simple and implicitly supports only the extract operation now (the following is identical on EL 8, Fedora 30, Fedora 31, Fedora 32)

```
#!/usr/bin/sh

#set -vx

# At this time, while this script is trivial, we ignore any parameters given.
# However, for backwards compatibility reasons, future versions of this script must 
# support the syntax "update-ca-trust extract" trigger the generation of output 
# files in $DEST.

DEST=/etc/pki/ca-trust/extracted

# Prevent p11-kit from reading user configuration files.
export P11_KIT_NO_USER_CONFIG=1

# OpenSSL PEM bundle that includes trust flags
# (BEGIN TRUSTED CERTIFICATE)
/usr/bin/p11-kit extract --format=openssl-bundle --filter=certificates --overwrite --comment $DEST/openssl/ca-bundle.trust.crt
/usr/bin/p11-kit extract --format=pem-bundle --filter=ca-anchors --overwrite --comment --purpose server-auth $DEST/pem/tls-ca-bundle.pem
/usr/bin/p11-kit extract --format=pem-bundle --filter=ca-anchors --overwrite --comment --purpose email $DEST/pem/email-ca-bundle.pem
/usr/bin/p11-kit extract --format=pem-bundle --filter=ca-anchors --overwrite --comment --purpose code-signing $DEST/pem/objsign-ca-bundle.pem
/usr/bin/p11-kit extract --format=java-cacerts --filter=ca-anchors --overwrite --purpose server-auth $DEST/java/cacerts
/usr/bin/p11-kit extract --format=edk2-cacerts --filter=ca-anchors --overwrite --purpose=server-auth $DEST/edk2/cacerts.bin
```

I validated my presumption that copying the ca root to either location has the same end result from a system trust standpoint. In either case, the new root shows up in the following files as extracted by the existing calls to update-ca-certificate:

```
$ grep -i warden -r /etc/pki /usr/share/pki
/etc/pki/ca-trust/extracted/openssl/ca-bundle.trust.crt:# Warden Proxy Local CA (hylfing)
/etc/pki/ca-trust/extracted/pem/objsign-ca-bundle.pem:# Warden Proxy Local CA (hylfing)
/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem:# Warden Proxy Local CA (hylfing)
/etc/pki/ca-trust/extracted/pem/email-ca-bundle.pem:# Warden Proxy Local CA (hylfing)
Binary file /etc/pki/ca-trust/extracted/edk2/cacerts.bin matches
Binary file /etc/pki/ca-trust/extracted/java/cacerts matches
```

Sorry about the long-winded reply here. Wanted to do some of my own research before I accept/rejected this change. I'm wondering if perhaps what was run up against was caching done by a browser such as Firefox given the following note on the Fedora docs page:

> While the Firefox browser is able to use an added certificate without executing update-ca-trust, it is recommended to run update-ca-trust after a CA change. Also note that browsers, such as Firefox, Epiphany, or Chromium, cache files, and you might need to clear the browser’s cache or restart your browser to load the current system certificates configuration.

I have experienced both Firefox picking up the custom CA root automatically on a clean VM where Warden was installed prior to Firefox ever having been launched, and also it not seeming to be picked up when installing Warden after Firefox had been run, the latter likely having been encountered on my part due to the mentioned caching which Firefox does.

If more technical details can be provided, I'll happily re-consider, but I'm going to close this out for now as it does not appear to result in any functional change.

PR #191, opened 2020-07-12
Update base context vcl

mrtuvn:


davidalger:
@mrtuvn I'll happily accept a PR that updates the /graphql to only forcibly exclude authorized requests (i.e. with a bearer specified) from the cache rather than all of them (which is what the M2 template VCL used to do), but given this is for local development the usefulness of having a health check probe and the grace mode is minimal at best. The headers that Magento removes from requests also should be maintained.

If you want to open a new PR to update just the /graphql logic, go for it. In the meantime, i'm going to go ahead and close this as the bulk of this PR is not something I plan to have baked into the default varnish image. You're more than welcome to mount a custom VCL into the container for use on your project, or use your own custom varnish image however.

davidalger:
@mrtuvn Quick update here: The varnish images now include the update to pass auth'd /graphql requests per latest 2.3 sample VCL along with a couple other minor changes in 3876fee99.

PR #196, opened 2020-07-23
Add detection of WSL when starting the debug container

LeeSaferite:
The current code for `warden debug` does not detect WSL based environments and instead identifies them as standard Linux. This becomes a problem when trying to configure the Xdebug remote connect address. The Windows version of Docker has the magic "host.docker.internal" hostname that the Linux version of Docker does not have. Additionally, the fallback method of remote IP detection fails to generate the correct IP address on a WSL setup. This test is the suggested way of detecting a WSL Linux system.

https://github.com/microsoft/WSL/issues/423#issuecomment-221627364


complexthings:
Windows, HA

davidalger:
@LeeSaferite When you tested this, was it on WSL1 or WSL2?

LeeSaferite:
This is on WSL1.

The release name contains `microsoft-standard` on WSL2.

davidalger:
Verified this doesn't change behavior on either macOS nor on Fedora 31. Looks good. Thanks for the contribution @LeeSaferite

PR #218, opened 2020-08-25
Fix not needed dot at the end of the command

ihor-sviziev:


davidalger:
Thanks for the contribution @ihor-sviziev

PR #228, opened 2020-09-10
Scan for partial env in WARDEN_HOME_DIR to allow customizations

k4emic:
This addition allows users to customize the docker-compose files by adding similarly-named files in the `~/.warden/environments...` directory.


ihor-sviziev (approved):

davidalger:
Thanks for the contribution @k4emic. This should be included in the next tagged release (no ETA on exactly when I'll tag it though, as I'm about ready to go on paternity leave for a couple weeks and not sure I'll have time to complete the 0.10.0 milestone before then)

PR #241, opened 2020-10-02
Add ability to define custom environment templates.

Skullsneeze:
By changing the `if ... fi` checks for environment initialization to `if ... elif ... fi` structures we can allow for a "generic" environment setup. This should make it easier for people to define their own environment templates.

Also added some rules to .gitignore to exclude any added environments, except the ones that already exist.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,12 @@
 /.sync
 /Icon?
 /.idea
+/environments/*
+!/environments/includes
+!/environments/laravel
+!/environments/local
+!/environments/magento1
+!/environments/magento2
+!/environments/shopware
+!/environments/symfony
+!/environments/wordpress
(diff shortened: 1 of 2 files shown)


davidalger:
This PR doesn't really align with the direction Warden is going, and it doesn't work as-is either (you'd get an invalid env type error, unless you specify a valid type, and adding the settings indiscriminately at the end will break the `local` env type). The incorrect assumption here is that any of the types lacking a condition will have a database, etc. One can already run `warden env-init foobar.test local` and create a custom per-project definition. There is also a somewhat related PR #228 which I intend to incorporate, so going ahead and closing this one out. Appreciate the shout out though!

Skullsneeze:
Thanks for getting back to me. We actually did look into the local env, and I think we can indeed work with that. Thanks for looking into the suggestion though!

PR #268, opened 2020-12-04
Updates MFTF configuration with notes about some common issues

Lunaetic:
Updates mftf.md with several notes regarding getting it running in a default `magento2` environment type.


Lunaetic:
Hopefully my formatting here is acceptable - I wasn't sure _where_ the information should go - if anyone has a better suggestion, just holler at me and I'll adjust.

docs/configuration/mftf.md
```diff
@@ -15,6 +15,83 @@ SELENIUM_HOST=selenium-hub
...
+```
+version: "3.5"
+services:
+  selenium-chrome:
+    hostname: ${WARDEN_ENV_NAME}_selenium-chrome
+    image: selenium/standalone-chrome${WARDEN_SELENIUM_DEBUG:-}:3.8.1
+    extra_hosts:
+      - ${TRAEFIK_SUBDOMAIN:-app}.${TRAEFIK_DOMAIN}:${TRAEFIK_ADDRESS:-0.0.0.0}
+    environment:
+      - START_XVFB=false
```
davidalger, line 33: @Lunaetic @lbajsarowicz Any reason this change couldn't/shouldn't be simply made in the Warden default environment spec for MFTF? Doesn't make sense to me to include a broken Selenium setup (even if it's not really our fault) and then have a doc saying how to override it with one that works.

docs/configuration/mftf.md
```diff
@@ -15,6 +15,83 @@ SELENIUM_HOST=selenium-hub
...
+
+Some MFTF tests that test uploading images for Products (such as `AdminSimpleProductImagesTest` - runnable with: `vendor/bin/mftf run:test AdminSimpleProductImagesTest`) will fail using the default configuration, due to `nginx` not having a large enough `client_max_body_size`, or the `selenium-chrome` tab crashing because `selenium-chrome` has run out of memory.
+
+The `client_max_body_size` issue will present itself when attempting the `bmp` upload test in `AdminSimpleProductImagesTest`; it can be remedied by increasing `nginx`'s `client_max_body_size`.
+
+To do this, create the file `.warden/nginx/conf.d/custom.conf`, and populate it with:
+
+```
+client_max_body_size 100M;
+```
```
davidalger, line 52: This should be embedded in the `/etc/nginx/nginx.conf` within the Nginx image. No reason users should need to set a custom Nginx setting here. Looking at the role I've used for Nginx in production, the default value I've used there has been 200m, so let's go with that.

docs/configuration/mftf.md
```diff
@@ -15,6 +15,83 @@ SELENIUM_HOST=selenium-hub
...
+
+To remedy, the `shm_size` for the `selenium-chrome` container needs to be increased from the default of 64M.
+
+Edit `.warden/warden-env.yml`, and `shm_size` as shown below:
+
+```
+version: "3.5"
+services:
+  selenium-chrome:
+    shm_size: '1gb'
```
davidalger, line 84: @Lunaetic @lbajsarowicz Is this shm_size fairly static or can it fluctuate? I.e. if we change the MFTM support to use chrome vs the spoke/hub (as it is currently) the shm_size should just be set in the defaults. If it's a fairly static need for this to be 1gb, then it can simply be used, otherwise it could be made a variable. Thoughts?

davidalger (commented):
Lots of good info here, thank you! Posed a number of questions inline with the changes. The driving force behind them is: Let's get these necessary changes into the base templates for Selenium support, unless there is a good reason not to, rather than making the setup require more manual effort to function.

Lunaetic:
@davidalger Gotcha - I'll update this PR with the suggested changes this week

docs/configuration/mftf.md
Lunaetic: 1gb was an arbitrary size I chose while working on it; variable works, thanks

Lunaetic: Created a variable, set to 128mb, which still allows the tests to run from my testing.

docs/configuration/mftf.md
Lunaetic: Clarification - when you say "embedded in the nginx.conf within the Nginx image"; do you mean "added to the docker.io image"?

Lunaetic: I've created the following images using wardenenv/nginx:1.16 and wardenenv/nginx:1.17 as bases that include this change in the nginx.conf files - I've tested them, however I'm not sure what the process is to update the containers in warden. Do they get pulled into the wardenenv/ repo, or is updating the base.nginx.yml file to point at them acceptable?

davidalger:
Closing in favor of #349

Issue #267, opened 2020-11-25
Xdebug 3 support


ihor-sviziev:
Xdebug 3 has been released and makes impressive performance improvements.
Let's add support of it? https://twitter.com/derickr/status/1331613630991900674?s=21

Den4ik:
@davidalger Unfortunately, IUS updates slower than remi and xdebug 3 still not provided there.
@ihor-sviziev Currently, I use my own builds and if you want you may try to check at fork https://github.com/magenius-team/warden/tree/main

ihor-sviziev:
@Den4ik I believe we need to have separate image with xdebug3, so it will continue working for everyone who used xdebug 2, but with additional option to switch to xdebug 3

Den4ik:
@ihor-sviziev good point

Den4ik:
> Also could you explain what is the benefit of using separate fork instead of just sending Pull Request to this repo?

I use remi package repository. If @davidalger approve moving from ius to remi I'll prepare PR

celorodovalho:
@Den4ik how can I use your fork?
I mean, I have installed on linux using homebrew, and currently I'm facing a problem regarding this issue https://youtrack.jetbrains.com/issue/WI-56500
I know I can update xdebug from remi to 2.9.7 version, however I don't know how to do that from warden source code.
So, I would like to know how can I switch to your fork at least?

Den4ik:
@celorodovalho look at https://docs.warden.dev/installing.html#alternative-installation and use fork url instead of this repo.
You may use `PHP_XDEBUG_3=1` flag at `.env` file for using xdebug3 version

davidalger:
[...]

PR #279, opened 2020-12-22
Add xdebug3 support as separate image type

Den4ik:
**Description**
This PR provide support of xdebug3 as separate images that can be configured in `.env` file by `PHP_XDEBUG_3=1` directive

**Related Issue**
#267


ihor-sviziev (approved):

ihor-sviziev:
@davidalger could you check it?

davidalger:
Looks really good. I'm going to build these and give it a run. Since it's opt-in, I'm not too concerned with it breaking anyone's workflows. Good call on adding a feature flag for it. Perhaps sometime it can be removed and this go Xdebug 3 by default once it's stable and accepted by many.

As Xdebug 3 is only available for PHP 7.2+ the image build tooling will need to be updated to only build the xdebug3 variant for 7.2 and higher. _I'm working on addressing this now_ It'd_ be simpler if the images were simply updated to install xdebug3 when php version supports it, but that would require everyone start using it right away, which we don't want here for now.

davidalger:
Updated the build script so it should pass muster when this is merged (not attempting to build xdebug3 images on older versions of PHP) and then ran this on a local builder:
```
export PHP_VERSION=7.4 \
  && ./images/scripts/build.sh --push php-fpm/xdebug3 \
  && ./images/scripts/build.sh --push php-fpm/magento1/xdebug3 \
  && ./images/scripts/build.sh --push php-fpm/magento2/xdebug3
```

The following images are now available on docker hub so this can be tested:

```
wardenenv/php-fpm:7.4-xdebug3
wardenenv/php-fpm:7.4-magento1-xdebug3
wardenenv/php-fpm:7.4-magento2-xdebug3
```

Once merged the rest should build via CI.

davidalger:
Spun up a 2.4.2 sandbox running PHP 7.4, added verified requests work and would stop on a breakpoint with and without `PHP_XDEBUG_3=1` using VSCode. Success! Going to merge this, and hopefully tag a release sometime this week. Thanks!

PR #280, opened 2020-12-29
Add zsh autocompletion definition

Lunaetic:
This PR adds an autocompletion definition script for zsh that uses sed and awk to parse the output of `usage.help`.

It also adds `sync` and `vnc` to the `usage.help` file, and corrects a typo.

### DIFF
--- a/autocompletion/zsh/warden.plugin.zsh
+++ b/autocompletion/zsh/warden.plugin.zsh
@@ -1,0 +1,20 @@
+#compdef warden
+
+# Installation
+# ============
+#
+# Copy this file, `warden.plugin.zsh`, to `~/.oh-my-zsh/plugins/custom/warden/`
+#
+# Enable the plugin by adding `warden` to `plugins=(… warden)` in `~/.zshrc`
+#
+# Reload your zsh session, or source `~/.zshrc`
+
+_warden_get_command_list() {
+    warden | sed -n '/Commands:/,$p' | sed '1d' | awk '{print $1}'
+}
+
+_warden() {
+    compadd `_warden_get_command_list`
+}
+
+compdef _warden warden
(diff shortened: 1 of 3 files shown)


sprankhub:
Thanks for a start on this! This is a very good start. In the end, it would be nice to have more advanced autocompletion for options or commands, which are "nested" like `warden env up` (currently, it would not complete the `up` if I understand correctly).

sprankhub:
Yes, this is still relevant.

sprankhub:
@davidalger, mind merging? There is much more work to do, but it should not stop us from merging this IMHO.

sprankhub:
@davidalger, mind having a look at this?

sprankhub:
@davidalger, should we just merge it and improve it whenever someone has time?

sprankhub:
I still think this should be merged :(

celorodovalho:
I think this is still relevant and should be merged.

navarr (changes requested):
Following the instructions did not work.  "plugin 'warden' not found"

Ideally, I'd like to do something similar to what Heroku does to install autocomplete scripts:

https://github.com/heroku/homebrew-brew/blob/master/Formula/heroku.rb

autocompletion/zsh/warden.plugin.zsh
```diff
@@ -0,0 +1,20 @@
+#compdef warden
+
+# Installation
+# ============
+#
+# Copy this file, `warden.plugin.zsh`, to `~/.oh-my-zsh/plugins/custom/warden/`
```
ihor-sviziev, line 6: maybe better to create a symlink? So if the file will update in next releases, it will be automatically updated there as well.

navarr:
Closing this as I don't foresee additional work occurring in this PR.  Myles, my man, feel free to reopen if you get a hankering to try a second stab at this.

Lunaetic:
@navarr 👍🏻 I'll take another stab in the coming weeks.

bap14:
@Lunaetic Another thing to consider is that you should start with the base zsh config, and then publish alternative installation for packages like oh-my-zsh. If you just source the completion file you have committed from `~/.zshrc` it does work, but as @navarr pointed out, it only does the top-level commands (env, env-init, db, etc.)

PR #288, opened 2021-01-18
[PHP-FPM] Fix cli debug auto start

Den4ik:
**Description**
This PR provides fix for autostart debug for cli commands and by a trigger for fpm

#267


Den4ik:
@davidalger Please check and merge fix for cli debug

davidalger:
Changes look good to me. Merging without testing since it only affects the xdebug3 images, so they can be rebuilt by CI. Would be good if you could pull and validate once they build.

ihor-sviziev:
@davidalger just checked - works fine

Issue #253, opened 2020-10-23
Support .modman environments


norgeindian:
For Magento1 projects we often use .modman and include modules via symlinks. 
So our project root folder has two subfolders: `.modman` and `shop` whereas most of the things in `shop` are symlinked from `modman`.
I firstly initialized warden in the shop folder, but then he does not include the `.modman` folder into the container, so can not use the symlinks.
I then tried to initialize warden in the project root folder. This way he has all the needed files, but of course does not know where to look for the Magento `index.php`.
When I add the `shop` folder as `WARDEN_WEB_ROOT` we have the same issue as in the beginning. The symlinked files are not included in the container.
Is there a way to fix that somehow?

norgeindian:
@davidalger , this would still be important for us.
Would be great if you could shortly explain, how I can manage to use warden for our environment structure.

PR #290, opened 2021-02-07
Add modman support for Magento1

norgeindian:
With this approach it would also be possible to handle modman environments in Magento1.
Solves #253


sprankhub:
Looks like an easy change. However, IMHO, we could also auto-detect modman-environments. If there is a `.modman` directory in the Warden root, we have a modman-environment. And in this case, `.modman/.basedir` already contains the necessary document root. Would IMHO be nicer to auto-detect this and not introduce another setting for this.

norgeindian:
@sprankhub , thanks for the idea.
Adapted my pull request, so modman environments are now automatically detected and handled.
What do you think about this approach?

commands/env.cmd
```diff
@@ -29,6 +29,12 @@ if [[ ${WARDEN_ENV_TYPE} != local ]]; then
     WARDEN_REDIS=${WARDEN_REDIS:-1}
 fi
 
+if [[ ${WARDEN_ENV_TYPE} == "magento1" && -f "${WARDEN_ENV_PATH}/.modman/.basedir" ]]; then
+  export WARDEN_MODMAN_ROOT='/'`cat "${WARDEN_ENV_PATH}/.modman/.basedir"`
```
sprankhub, line 33: I am not a bash expert at all, but https://www.shellcheck.net/ recommends to [not use backticks](https://github.com/koalaman/shellcheck/wiki/SC2006). `'/'$(cat "${WARDEN_ENV_PATH}/.modman/.basedir")` should be a proper alternative.

commands/env.cmd
```diff
@@ -29,6 +29,12 @@ if [[ ${WARDEN_ENV_TYPE} != local ]]; then
...
+else
+  export WARDEN_MODMAN_ROOT='/'
```
sprankhub, line 35: I think the name is weird. Yes, we use it for a `modman`-specific fix, but it is used in all cases.

1. Maybe we can directly set it to `NGINX_PUBLIC`?
2. If not, maybe we use a generic name like `NGINX_PUB_DIR`?

commands/env.cmd
norgeindian: Thanks, adpated

commands/env.cmd
norgeindian: Directly is not possible, at least I did not find a way.
So I find this naming already better, you're right

sprankhub:
Nice! Let's see what @davidalger says ;-)

davidalger (commented):
Sounds like a great idea to me. Left a couple comments / nit-picks. Once those are addressed, be happy to merge it. Thanks!

norgeindian:
@davidalger , thanks a lot for your comments and suggestions.
I included them, so feel free to merge.

Issue #289, opened 2021-02-03
Command history cleans up all the time


ihor-sviziev:
Hi,

I found that when I'm rebooting my computer and then running `warden env up` - it all the time re-creates php-fpm container. As a result, the entire command history in the `warden shell` is getting lost.

It would be great if we could persist command history OR do not re-create the php-fpm container when it's not really needed (for instance new image was pulled, or some config has changed, etc.)

Den4ik:
@ihor-sviziev if you will use `warden env start` container will not be recreated

sprankhub:
+1 for persisting the command history.

ihor-sviziev:
@Den4ik looks like `warden env up` re-creates php-fpm container because of changing the `SSH_AUTH_SOCK` variable after reboot (I see this behavior on WSL2 + [the following solution for ssh agent](https://esc.sh/blog/ssh-agent-windows10-wsl2/).

It seems if I'll switch to the solution with single static ssh auth sock - it shouldn't re-create containers.
```bash
if ! pgrep ssh-agent > /dev/null; then
  rm -f /tmp/ssh-auth-sock
  eval "$(ssh-agent -s -a /tmp/ssh-auth-sock)"
  ssh-add
else
  export SSH_AUTH_SOCK=/tmp/ssh-auth-sock
fi
```

davidalger:
@ihor-sviziev Interesting workaround. That would explain why I hadn't experienced the behavior myself, primary driver being Mac OS only occasionally working on my Linux box. Using start vs up would prevent the re-create as well then, since it docker-compose doesn't look for changes in the container spec when only start has been requested. Sounds like we've got the best of both worlds: workaround to prevent re-create _and_ persisted bash history thanks to the PR from @Den4ik making it happen.

Den4ik:
@ihor-sviziev interesting. I didn't work on windows 10 years, but for now I really want to check it ))

ihor-sviziev:
One interesting thing - today I had the same issue on Ubuntu 20.04, but with known hosts. After reboot - the php-fpm container was re-created and the `~/.ssh/known_hosts` file was removed, so I wanted to check why that container was re-created.

I checked what was the output of `warden env config` before and after reboot - diff is interesting - that's IP of the traefik in the extra_hosts section:
https://github.com/davidalger/warden/blob/3cc14a24867191342564748962375a3d5d3c879b/environments/includes/php-fpm.base.yml#L10-L11

[...]

PR #304, opened 2021-02-15
[PHP-FPM] Mount bash history as separate volume to prevent cleaning after recreation

Den4ik:
**Resolver Issue:**
#289 Command history cleans up all the time


navarr (approved):

davidalger:
Thanks for the contribution @Den4ik

PR #308, opened 2021-02-19
Fixed bash history folder permissions

Den4ik:
Related PR #304 

I faced with folder permissions that prevent history file creation.
This PR provides fix for bash history folder permissions.

@davidalger I'm not sure that it's best solution, but currently, I don't have any ideas.

### DIFF
--- a/commands/env.cmd
+++ b/commands/env.cmd
@@ -51,6 +51,8 @@ appendEnvPartialIfExists "networks"
 
 if [[ ${WARDEN_ENV_TYPE} != local ]]; then
     appendEnvPartialIfExists "php-fpm"
+    # define bash history folder for changing permissions
+    export CHOWN_DIR_LIST=/var/bash_history
 fi
 
 [[ ${WARDEN_NGINX} -eq 1 ]] \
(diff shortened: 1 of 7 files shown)


davidalger:
This is a working solution on Mac OS environments, but doesn't apply to Linux (which will suffer the same issue, since it's due to Docker volumes being created as root:root no matter what).

Side effect (not sure this is a big deal really given these are dev-only containers) is `/var` is also chown'd since the CHOWN_DIR_LIST loop walks up the path:
```
$ ls -lhd /var
drwxr-xr-x 1 www-data www-data 4.0K Feb 23 04:37 /var
```

The only alternate solution I can think of would be a modification to the entrypoint script, but that would also require any one updating Warden to pull latest images (or suffer having bash history stop working between `warden shell` invocations).

I think the side-effect of using CHOWN_DIR_LIST is fine, so let's add `CHOWN_DIR_LIST=${CHOWN_DIR_LIST:-}` to the `php-fpm.base.yml` as well so the exported value will be included on Linux configurations as well.

Den4ik:
> The only alternate solution I can think of would be a modification to the entrypoint script, but that would also require any one updating Warden to pull latest images (or suffer having bash history stop working between `warden shell` invocations).

Agree. I thought about it, but as you noted it requires to rebuild images and pull the latest from docker hub

Den4ik:
@davidalger I have moved `CHOWN_DIR_LIST` into `php-fpm.base.yml` and mount into `/bash_history` instead of `/var/bash_history`
Also, I added `WARDEN_CHOWN_DIR_LIST ` variable that allows defining custom `chown` list at project `.env` file

PR #309, opened 2021-02-19
[XDEBUG3] Enable only debug for cli command

Den4ik:
When I used xdebug at cli docker volume disk space has run out due profiler and tracer logs.

This PR enable only debug mode for cli container.


davidalger:
Good call!

PR #310, opened 2021-02-19
Add warden db dump command

Den4ik:
I use this command when need fastly check the issue on environment DB but don't want to lose local

### DIFF
--- a/commands/db.cmd
+++ b/commands/db.cmd
@@ -39,6 +39,10 @@ case "${WARDEN_PARAMS[0]}" in
             | "${WARDEN_DIR}/bin/warden" env exec -T db \
             mysql -u"${MYSQL_USER}" -p"${MYSQL_PASSWORD}" --database="${MYSQL_DATABASE}" "${WARDEN_PARAMS[@]:1}" "$@"
         ;;
+    dump)
+            "${WARDEN_DIR}/bin/warden" env exec -T db \
+            mysqldump -u"${MYSQL_USER}" -p"${MYSQL_PASSWORD}" "${MYSQL_DATABASE}" | pv | gzip > "${WARDEN_PARAMS[@]:1}" "$@"
+        ;;
     *)
         fatal "The command \"${WARDEN_PARAMS[0]}\" does not exist. Please use --help for usage."
         ;;
--- a/commands/db.cmd
+++ b/commands/db.cmd
@@ -41,7 +41,7 @@ case "${WARDEN_PARAMS[0]}" in
         ;;
     dump)
             "${WARDEN_DIR}/bin/warden" env exec -T db \
-            mysqldump -u"${MYSQL_USER}" -p"${MYSQL_PASSWORD}" "${MYSQL_DATABASE}" | pv | gzip > "${WARDEN_PARAMS[@]:1}" "$@"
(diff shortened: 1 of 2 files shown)


commands/db.cmd
```diff
@@ -39,6 +39,10 @@ case "${WARDEN_PARAMS[0]}" in
             | "${WARDEN_DIR}/bin/warden" env exec -T db \
             mysql -u"${MYSQL_USER}" -p"${MYSQL_PASSWORD}" --database="${MYSQL_DATABASE}" "${WARDEN_PARAMS[@]:1}" "$@"
         ;;
+    dump)
+            "${WARDEN_DIR}/bin/warden" env exec -T db \
+            mysqldump -u"${MYSQL_USER}" -p"${MYSQL_PASSWORD}" "${MYSQL_DATABASE}" | pv | gzip > "${WARDEN_PARAMS[@]:1}" "$@"
```
ihor-sviziev, line 44: 1. I believe we should stip all the "definer" (and maybe some global things) we're doing in the import command. What do you think?
2. I think using gzip should be optional, maybe with some flag? I think would be good to have support for flow like this: 
```bash
warden db dump > dump.sql 
pv dump.sql | warden db import
```

ihor-sviziev (changes requested):

commands/db.cmd
Den4ik: Hi @ihor-sviziev 
Thanks, good suggestion

davidalger: Let's have it output vanilla mysldump results if we can, as this will allow the dump to be inspected to see what was actually inside the db container. The piping here makes gzip and pv become dependencies (gzip is typically present by default on workstations, but pv often must be installed by a user). 

```suggestion
            mysqldump -u"${MYSQL_USER}" -p"${MYSQL_PASSWORD}" "${MYSQL_DATABASE}" "${WARDEN_PARAMS[@]:1}" "$@"
```

davidalger: As written, it causes errors, so this would need to be updated regardless.
```
davidalger@silver-bullet:~/Sites/m2sandbox (develop) $ warden db dump
/opt/warden/commands/db.cmd: line 44: "${WARDEN_PARAMS[@]:1}": ambiguous redirect
0.00 B 0:00:01 [0.00 B/s] [<=>                                                                                                                                                ]
ERROR: Command `gzip "$@" > "${WARDEN_PARAMS[@]:1}"` at /opt/warden/commands/db.cmd:31 failed with exit code 1
```

Verified the suggested change here functions using `warden db dump` (prints to shell) or `warden db dump | pv | gzip > test.sql.gz` (dumps compressed dump to file with status)

commands/db.help
```diff
@@ -9,7 +9,8 @@ WARDEN_USAGE=$(cat <<EOF
   -h, --help        Display this help menu
 
 \033[33mCommands:\033[0m
-  connect           Launches an interactive mysql session within the current project environment
-  import            Reads data from stdin and loads it into the current project\047s mysql database
+  connect                  Launches an interactive mysql session within the current project environment
+  import                   Reads data from stdin and loads it into the current project\047s mysql database
+  dump <dump_file.sql.gz>  Create database dump and put it into specified file
```
davidalger, line 14:
```suggestion
  dump                     Create database dump and put it into specified file
```

ihor-sviziev (approved):
Looks good to me.

PR #317, opened 2021-03-09
Sample for building arm64 images with GH actions

k4emic:
With the rise of arm64 on Apple products, we've experienced that not all warden images are playing equally nice.

Based on some early feedback from some users at Vaimo, some images have problems running through the built-in x86 emulator.

Namely
- elasticsearch -> Platform-specific JVM
- varnish -> No official arm support
- redis -> qemu doesn't play nicely "qemu: uncaught target signal 11 (Segmentation fault) - core dumped"

This pull request (in its current form) is only a placeholder to facilitate a discussion on how (and if) Warden should approach providing support for Apple's new M1 silicon.

In this PR, I've simply added buildx support and default to building for both linux/amd64 and linux/arm64 architectures. 

Running this locally with `nektos/act` works (at least on my mid '15 macbook), but I'm no expert on this subject.

### DIFF
--- a/.github/workflows/docker-image-elasticsearch.yml
+++ b/.github/workflows/docker-image-elasticsearch.yml
@@ -15,6 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1
+    - uses: docker/setup-buildx-action@v1
     - run: ./images/scripts/build.sh --push "${BUILD_GROUP}"
       env:
         BUILD_GROUP: elasticsearch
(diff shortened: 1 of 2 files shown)


davidalger:
@k4emic Now that I've got #345 wrapped up, I'll begin to give this a go as I have the time to do so. Rebuilding all the images may or may not readily work, as for example, to my knowledge Varnish packages for ARM don't yet exist.

At the moment I'll have to rely on others to test the ARM variants of the images though, once built, as I don't currently an M1 chip to test them on. All of my MacOS and Linux machines are running Intel chipsets.

I may use CI based builds pushing into a temporary location for purposes of initially testing the dual-platform builds to avoid potentially causing problems with the existing images until the updated build process is known to work and create functional images.

drpayyne:
I might have something brewing in this space. Will get back soon! ☕

davidalger:
Thanks to both of you for the efforts here. Will be addressing multi-arch builds in the manner implemented on #505

PR #334, opened 2021-03-25
Use correct socket path for linux

### LINKED ISSUE #331: SSH Agent freezes sometimes
I discovered that sometimes I was not able to clone git repositories configured in composer.json as type VCS.
I was freezing.
After a lot of debug I discovered that the issue is caused by the socat socket proxy.

## Steps to reproduce 

Run `ssh-add -l` several times, fast enough (as composer is doing during composer install)

Output from `/var/log/socat-ssh-auth.log`

```
2021/03/19 20:50:41 socat[2992] E write(7, 0x5642d4464e70, 5): Broken pipe
2021/03/19 21:00:29 socat[3018] E write(7, 0x5642d4464e70, 887): Broken pipe
```

This can't be reproduced as `root` and set `/run/host-services/ssh-auth.sock` as `SSH_AUTH_SOCK`

Reproduced on MacOS and Ubuntu 20

alinalexandru:
Fixes #331

I explained in ticket the reason to set the `SSH_AUTH_SOCK` environment variable

### DIFF
--- a/commands/env.cmd
+++ b/commands/env.cmd
@@ -48,6 +48,11 @@ if [[ ${XDEBUG_CONNECT_BACK_HOST} == '' ]] && grep -sqi microsoft /proc/sys/kern
     export XDEBUG_CONNECT_BACK_HOST=host.docker.internal
 fi
 
+## For linux, if UID is 1000, there is no need to use the socat proxy.
+if [[ ${WARDEN_ENV_SUBT} == "linux" && $UID == 1000 ]]; then
+    export SSH_AUTH_SOCK_PATH_ENV=/run/host-services/ssh-auth.sock
+fi
+
 ## configure docker-compose files
 DOCKER_COMPOSE_ARGS=()
 
--- a/environments/includes/blackfire.base.yml
+++ b/environments/includes/blackfire.base.yml
@@ -16,7 +16,7 @@ services:
     environment:
       - TRAEFIK_DOMAIN
       - TRAEFIK_SUBDOMAIN
-      - SSH_AUTH_SOCK=/tmp/ssh-auth.sock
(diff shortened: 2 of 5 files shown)


ihor-sviziev:
Hi @alinalexandru,
I believe it won't work correctly in case if your user id on the host machine not equal to 1000. 
Here are more details about that: https://github.com/davidalger/warden/pull/307#issuecomment-781921198 https://github.com/davidalger/warden/pull/307#issuecomment-783863500

alinalexandru:
You are right with the user id.
The issue with user with ID 1000 still exists. This week I had to change the user ID to 1000 so that I could fix the `composer/cache` permissions.

Don't know if there could be done a user id check in entry point.
For us at the moment, this hack is working, as we all have User ID 1000.

ihor-sviziev:
@alinalexandru maybe we can introduce a new variable that will be set similar to this?
https://github.com/davidalger/warden/blob/5d1843bca3f6ed0ebc8615d651db4bc57765bb91/commands/env.cmd#L46-L49

alinalexandru:
That it's a good ideea, it might solve this issue for most of the linux users.

ihor-sviziev (changes requested):
Could you update your solution with suggested in comments changes?

alinalexandru:
Yes, I will work on that today.

ihor-sviziev (approved):

springerin:
Any updates on when this can be release? 

Great job @alinalexandru for the debug. Thank you very much, i am having this issue very often now and couldn't figure out what was happening.

ihor-sviziev:
@davidalger could you review this PR?

davidalger:
LG2M as well. Thanks for the efforts at guiding this one @ihor-sviziev. Going to merge it, and once I have a chance to verify on my Linux box over here and update the CL will tag a release.

PR #343, opened 2021-04-05
Laravel: increase read timeout to prevent 504s when debugging

dmitriyklyuzov:
When debugging a Laravel project, the site times out after 60 seconds because the `fastcgi_read_timeout` param is set to a low default value. Since this is a development environment, I am increasing this value to 3200s to match Magento configuration.

### DIFF
--- a/images/nginx/context/etc/nginx/available.d/application.conf
+++ b/images/nginx/context/etc/nginx/available.d/application.conf
@@ -13,7 +13,7 @@ location ~ \.php$ {
     fastcgi_buffers 1024 4k;
     fastcgi_buffer_size 32k;
     fastcgi_busy_buffers_size 256k;
-    fastcgi_read_timeout 60s;
+    fastcgi_read_timeout 3200s;
 
     include fastcgi_params;
 
--- a/images/nginx/context/etc/nginx/available.d/application.conf
+++ b/images/nginx/context/etc/nginx/available.d/application.conf
@@ -13,7 +13,7 @@ location ~ \.php$ {
     fastcgi_buffers 1024 4k;
     fastcgi_buffer_size 32k;
     fastcgi_busy_buffers_size 256k;
-    fastcgi_read_timeout 3200s;
+    fastcgi_read_timeout 600s;
 
     include fastcgi_params;
 


images/nginx/context/etc/nginx/available.d/application.conf
```diff
@@ -13,7 +13,7 @@ location ~ \.php$ {
     fastcgi_buffers 1024 4k;
     fastcgi_buffer_size 32k;
     fastcgi_busy_buffers_size 256k;
-    fastcgi_read_timeout 60s;
+    fastcgi_read_timeout 3200s;
```
ihor-sviziev, line 16: I just checked - this param is set to 600s in Magento2 installation. 10 minutes actually more than enough for debugging purposes. As Magento 1 based config is obviously more older than for Magento 2 i would suggest to replace this param in all places to the following:
```suggestion
    fastcgi_read_timeout 600s;
```

What do you think about that?

ihor-sviziev (changes requested):

ihor-sviziev (approved):

davidalger:
Thanks for the contribution @dmitriyklyuzov. Accepted the change with `600s` to be in line with Magento 2 config (Magento 1 is up at 3200s for different reasons).

Images have built and should be available, you'll simply need to pull and re-create with them:

```
warden env pull && warden env up
```

dmitriyklyuzov:
@davidalger @ihor-sviziev I have no problem setting this to a lower limit of 600 seconds, but what about this entry in magento2.conf? Same variable is set to 3600s there (I mistakenly originally set this to 3200 instead of 3600)

https://github.com/davidalger/warden/blob/8d047d18fb41d69e2366bebb100a982c4a95118f/images/nginx/context/etc/nginx/available.d/magento2.conf#L152

dmitriyklyuzov:
> Thanks for the contribution @dmitriyklyuzov. Accepted the change with `600s` to be in line with Magento 2 config (Magento 1 is up at 3200s for different reasons).
> 
> Images have built and should be available, you'll simply need to pull and re-create with them:
> 
> ```
> warden env pull && warden env up
> ```

Understood, thanks!

ihor-sviziev:
I think we should set this value to 600s in all places (except Magento 1,
if there is some use case for that)

davidalger:
The higher values on timeouts such as 3200 and 3600 are due to these configurations ultimately coming from production configurations, with development related things added to them, I didn't actually check to verify M2 was set at 600s but the reason for those being higher are Magento admin requests which take a _long_ time, and so to avoid 502 responses the timeouts were raised in the std configurations I used for production deployments. There probably really is no reason any of them need to be that high on a local development machine.

PR #358, opened 2021-04-21
docs: Add installation instructions via Copr repository

tdgroot:
To easen installation and maintenance of Warden on Fedora systems, I initiated a Copr repository.

For now the spec file is hosted on a separate [Git repository on Pagure.io](https://pagure.io/warden-copr), if you'd like to host the RPM spec file in the main Warden repository, let me know :).

**EDIT**
The Copr package now also places the `systemd-resolved` file (as described by #360) and restarts `systemd-resolved` after installing/upgrading.


ihor-sviziev:
Great idea! Would like to have similar thing for Ubuntu

ihor-sviziev:
@davidalger, what do you think?

PR #360, opened 2021-04-22
docs: Update docs about DNS resolution using systemd-resolved

tdgroot:
Usage of systemd-resolved is the most simple and out of the box configuration for any system that is shipped with systemd + systemd-resolved. So it shouldn't be Ubuntu specific!


sprankhub (approved):

davidalger:
Very nice. Thank you!

ihor-sviziev:
I believe, we can do that automatically during setting up warden, right?

PR #364, opened 2021-04-23
Add M1 multiple store explanation

### LINKED ISSUE #231: Run Multiple Website Scopes From Sub Folders
Hi - Is it possible to run multiple website scopes from sub folders using warden? For example:

https://app.example.test/uk -> Points to UK website
https://app.example.test/us -> Points to US website

With this setup all category and product pages return 404s and the US website seems to redirect to the UK website e.g. https://app.example.test/us/product-url -> https://app.example.test/uk/us/product-url.

I have the subfolders created correctly as I have this working on a live environment. When debugging it seems that all requests are routed to pub/index.php rather than the subfolders index.php which I think is what is causing the issue.

Does anyone know how to get this kind of setup to work using Warden?

norgeindian:
Closes #231

### DIFF
--- a/docs/configuration/multipledomains.md
+++ b/docs/configuration/multipledomains.md
@@ -64,6 +64,92 @@ Multiple top-level domains may also be setup by following the instructions below
          - sub1.alternate2.test:${TRAEFIK_ADDRESS:-0.0.0.0}
          - sub2.alternate2.test:${TRAEFIK_ADDRESS:-0.0.0.0}
     ```
+### Magento 1 Subfolder setup
+If you have a M1 subfolder setup to run your multiple stores, you will have to extend the `nginx` configuration. To do that, add `.warden/warden-env.yml`
+file with the following content to your project root:
+```
+version: "3.5"
+services:
+  nginx:
+    volumes:
+      - ./.warden/nginx/custom.conf:/etc/nginx/default.d/custom.conf
+``` 
+Now you add a file `.warden/nginx/custom.conf` in your project root with the following content:
+```
+location /subdomain1 {
+    alias /var/www/html/;
+    try_files $uri $uri/ @subdomain1;
+
+   location ~ \.php$ {
+       try_files $uri =404;
+       expires off;
+
+       fastcgi_pass $fastcgi_backend;
+
+       fastcgi_buffers 1024 4k;
+       fastcgi_buffer_size 128k;
+       fastcgi_busy_buffers_size 256k;
+       fastcgi_read_timeout 3600s;
+
+       include fastcgi_params;
+
+       fastcgi_param HTTPS on;
+
+       # Prevents these headers being used to exploit Zend_Controller_Request_Http
+       fastcgi_param HTTP_X_REWRITE_URL "";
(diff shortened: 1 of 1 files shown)


docs/configuration/multipledomains.md
```diff
@@ -64,6 +64,92 @@ Multiple top-level domains may also be setup by following the instructions below
...
+```
+location /subdomain1 {
+    alias /var/www/html/;
+    try_files $uri $uri/ @subdomain1;
+
+   location ~ \.php$ {
+       try_files $uri =404;
+       expires off;
+
+       fastcgi_pass $fastcgi_backend;
```
sprankhub, line 87: I am not really into nginx, but to me, it looks weird that we have to add all these configuration options from the default Magento nginx configuration again. Maybe there is a better way? In a perfect world, we would only add `MAGE_RUN_CODE` here.

docs/configuration/multipledomains.md
```diff
@@ -64,6 +64,92 @@ Multiple top-level domains may also be setup by following the instructions below
...
     ```
+### Magento 1 Subfolder setup
+If you have a M1 subfolder setup to run your multiple stores, you will have to extend the `nginx` configuration. To do that, add `.warden/warden-env.yml`
+file with the following content to your project root:
+```
+version: "3.5"
+services:
+  nginx:
+    volumes:
+      - ./.warden/nginx/custom.conf:/etc/nginx/default.d/custom.conf
```
sprankhub, line 75: Are you sure this works with `custom.conf`? Aren't the config files read alphabetically, so that we need to make sure that the custom file is read after `magento1.conf`?

docs/configuration/multipledomains.md
norgeindian: I tried to remove the other parts, but that always failed.
Maybe someone else has a deeper knowledge of nginx config and could help here?

docs/configuration/multipledomains.md
norgeindian: I tried it, and it worked, but a good point anyway.
Just checked `images/nginx/context/etc/nginx/conf.d/default.conf` again and found the following:

```
    include /etc/nginx/available.d/${NGINX_TEMPLATE};
    include /etc/nginx/default.d/*.conf;
```
So we're definitely loaded after `magento1.conf`, so this should be fine

tdgroot: @sprankhub the application config files are loaded from `/etc/nginx/available.d`, before loading files from `/etc/nginx/conf.d`.

docs/configuration/multipledomains.md
tdgroot: It is possible to simply include an nginx config file to handle the fastcgi stuff. Also very common is to use `echo_exec` to pass the request to php-fpm, which looks very clean.

sprankhub:
@davidalger, mind reviewing?

sprankhub:
@davidalger...?

sprankhub:
Still relevant.

sprankhub:
Still relevant. @davidalger, stale bot on PRs is really not a good idea :(

sprankhub:
Yes, it is :angry:

PR #384, opened 2021-06-02
Downgrade to mysql 8.0.23

sebastian-ehrling:
Currently there is a problem with running Magento from mysql 8.0.24 and higher.
See changes on collations in mysql 8.0.24 -> https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-24.html

### DIFF
--- a/images/mysql/8.0/Dockerfile
+++ b/images/mysql/8.0/Dockerfile
@@ -1,1 +1,1 @@
-FROM mysql:8.0
+FROM mysql:8.0.23


davidalger:
The only reference to changes that pertain to collations in those release notes don't come across as backwards incompatible:

> Important Note: When a utf8mb3 collation was specified in a CREATE TABLE statement, SHOW CREATE TABLE, DEFAULT CHARSET, the values of system variables containing character set names, and the binary log all subsequently displayed the character set as utf8 which is becoming a synonym for utf8mb4. Now in such cases, utf8mb3 is shown instead, and CREATE TABLE raises the warning 'collation_name' is a collation of the deprecated character set UTF8MB3. Please consider using UTF8MB4 with an appropriate collation instead. (Bug #27225287, Bug #32085357, Bug #32122844)


> Selecting from the Performance Schema replication_group_members table in some cases raised Error 1267 Illegal mix of collations .... This was due to a change in MySQL 8.0.22 in the handling of comparisons between a column value and a system variable. (Bug #32244631)

Magento officially supports MySQL 8.0 as of 2.4, but I can't find anything describing a compatibility issue caused by this patch release to MySQL. Beyond that though, I have no intention of freezing a generic database container to a specific minor release, as that will defeat the auto-build that occurs each month to keep them updated.

sebastian-ehrling:
ok thx!

PR #385, opened 2021-06-02
Update db.base.yml

### LINKED ISSUE #354: Using MySQL 8 with magento2
Hi @davidalger 

with magento2 and your mysql8 image (docker.io/wardenenv/mysql:8.0) there is an error when importing existing magento databases.
The following setting resolves the problem. Is it possible to set this setting by default in your mysql8 image?

[mysqld]
log_bin_trust_function_creators=ON

sebastian-ehrling:
solved following error when importing a dump:

ERROR 1419 (HY000) at line 693: You do not have the SUPER privilege and binary logging is enabled (you *might* want to use the less safe log_bin_trust_function_creators variable)

SEE here:
https://github.com/davidalger/warden/issues/354

close #354

### DIFF
--- a/environments/magento2/db.base.yml
+++ b/environments/magento2/db.base.yml
@@ -10,3 +10,4 @@ services:
       - mysqld
       - --max_allowed_packet=1024M
       - --explicit_defaults_for_timestamp=on
+      - --log_bin_trust_function_creators=on


ihor-sviziev (approved):

davidalger:
Closing in favor of #403 which makes this setting no longer necessary as it only applies to MySQL where binary logging is enabled..

PR #395, opened 2021-06-23
Add Blackfire CLI to Magento blackfire images

ihor-sviziev:
The BlackfireCLI was added in https://github.com/davidalger/warden/pull/188, but it doesn't work for Magento 2:
```bash
$ warden blackfire
OCI runtime exec failed: exec failed: container_linux.go:380: starting container process caused: exec: "blackfire": executable file not found in $PATH: unknown
```

I found that it Blackfire CLI was added to base php-fpm image, but not to magento-specific images. This PR fixes this issue.

**Workaround** (while this PR still not merged):
You can run add backfire CLI manually by executing the following command:
```bash
warden env exec --user root php-blackfire bash -c 'mkdir -p /tmp/blackfire && curl -L https://blackfire.io/api/v1/releases/client/linux_static/amd64 | tar zxp -C /tmp/blackfire && mv /tmp/blackfire/blackfire /usr/bin/blackfire && rm -rf /tmp/blackfire'
```

### DIFF
--- a/images/php-fpm/magento1/blackfire/Dockerfile
+++ b/images/php-fpm/magento1/blackfire/Dockerfile
@@ -8,6 +8,12 @@ RUN curl -o - "http://packages.blackfire.io/fedora/blackfire.repo" \
     && dnf clean all \
     && rm -rf /var/cache/dnf
 
+# Install the Blackfire Client to provide access to the CLI tool
+RUN mkdir -p /tmp/blackfire \
+    && curl -L https://blackfire.io/api/v1/releases/client/linux_static/amd64 | tar zxp -C /tmp/blackfire \
+    && mv /tmp/blackfire/blackfire /usr/bin/blackfire \
+    && rm -rf /tmp/blackfire
+
 COPY blackfire/etc/php.d/*.ini /etc/php.d/
 
 USER www-data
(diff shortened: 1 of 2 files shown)


davidalger:
I'm not sure this is the issue, because `warden blackfire` is expected to exec the CLI command within the `php-blackfire` container, which runs separately from the magento2 image: https://github.com/davidalger/warden/pull/188/files#diff-79d55b7b425f7730cd1e8c3f88a29c44ea69f75d39aa2b8f0c2817ab37e538b9R10

It sounds based on `executable file not found in $PATH: unknown` part of the error, that this may have been caused by the upgrade to EL 8 base images, after which there was an issue with one other thing an entrypoint called without using the fully qualified path to the executable.

ihor-sviziev:
@davidalger, 
We have 2 containers, one of them is blackfire-agent, 2nd one php-fpm container with added backfire PHP module (and supposed to contain backfire CLI).
https://github.com/davidalger/warden/blob/2f8c50cce1e551ddb7a50bfca0c7477e8c20538c/environments/includes/blackfire.base.yml#L12-L37

The `warden env config` now shows following:
```yml
services:
  blackfire-agent:
    ...
    image: blackfire/blackfire:latest
  php-blackfire:
    ...
    image: docker.io/wardenenv/php-fpm:7.3-magento2-blackfire
    ...
```

Or, you mean that we should run this command on the blackfire-agent container?

Den4ik (approved):

davidalger:
@ihor-sviziev I think I see what you're saying now. When I first looked at this, for some reason I thought this was adding the CLI command to the `php-fpm` container rather than the `php-blackfire` container images layered on top of m1/m2 base images. This makes a whole lot more sense to me this morning. :D

PR #397, opened 2021-06-28
Fixed php fpm cli memory limit | Removed unused context files

Den4ik:
By config from `01-php.ini` memory limit is `2G` for all images.

This PR provide next updates:
1. memory limits splits to cli and fpm configs
2. 756M of memory limit set by default for fpm
3. non memory limits set for cli
4. removed unused context files due to `build.sh` logic


Den4ik:
@davidalger Could you check this PR?

davidalger:
This looks good, thank you!

PR #413, opened 2021-08-18
Added redis shorthand command

werfu:
I was looking for a quick way to flush redis cache, so I added a shorthand to run redis cli or command directly.

### DIFF
--- a/commands/redis.cmd
+++ b/commands/redis.cmd
@@ -23,8 +23,4 @@ if [[ "${WARDEN_PARAMS[0]}" == "stat" ]]; then
   "${WARDEN_DIR}/bin/warden" env exec redis sh -c "redis-cli --stat"
 fi
 
-if (( ${#WARDEN_PARAMS[@]} == 0 )); then
-    "${WARDEN_DIR}/bin/warden" env exec redis sh -c "redis-cli"
-else
-    "${WARDEN_DIR}/bin/warden" env exec -T redis sh -c "redis-cli ${WARDEN_PARAMS}"
-fi
+"${WARDEN_DIR}/bin/warden" env exec redis redis-cli "${WARDEN_PARAMS[@]}" "$@"
--- a/commands/redis.cmd
+++ b/commands/redis.cmd
@@ -16,7 +16,11 @@ fi
 ## load connection information for the mysql service
 REDIS_CONTAINER=$(warden env ps -q redis)
 if [[ ! ${REDIS_CONTAINER} ]]; then
-    fatal "No container found for db service."
+    fatal "No container found for redis service."
+fi
+
+if [[ "${WARDEN_PARAMS[0]}" == "stat" ]]; then
+  "${WARDEN_DIR}/bin/warden" env exec redis sh -c "redis-cli --stat"
 fi
 
 if (( ${#WARDEN_PARAMS[@]} == 0 )); then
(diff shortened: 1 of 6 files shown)


commands/redis.cmd
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+[[ ! ${WARDEN_DIR} ]] && >&2 echo -e "\033[31mThis script is not intended to be run directly!\033[0m" && exit 1
+
+WARDEN_ENV_PATH="$(locateEnvPath)" || exit $?
+loadEnvConfig "${WARDEN_ENV_PATH}" || exit $?
+assertDockerRunning
+
+if [[ ${WARDEN_REIDS:-1} -eq 0 ]]; then
+  fatal "Redis environment is not used (WARDEN_REDIS=1)."
+fi
```
davidalger, line 10:
```suggestion
if [[ ${WARDEN_REDIS:-1} -eq 0 ]]; then
  fatal "Redis environment is not used (WARDEN_REDIS=0)."
fi
```

commands/redis.cmd
```diff
@@ -0,0 +1,30 @@
...
+
+if [[ "${WARDEN_PARAMS[0]}" == "stat" ]]; then
+  "${WARDEN_DIR}/bin/warden" env exec redis sh -c "redis-cli --stat"
+fi
+
+if (( ${#WARDEN_PARAMS[@]} == 0 )); then
+    "${WARDEN_DIR}/bin/warden" env exec redis sh -c "redis-cli"
+else
+    "${WARDEN_DIR}/bin/warden" env exec -T redis sh -c "redis-cli ${WARDEN_PARAMS}"
+fi
```
davidalger, line 30:
```suggestion
"${WARDEN_DIR}/bin/warden" env exec redis redis-cli "${WARDEN_PARAMS[@]}" "$@"

```

davidalger: This will support all of the following cases:

* `warden redis` - launch interactive prompt
* `warden redis info` - show all-inclusive info output
* `warden redis info keyspace` - show keyspace info only
* `warden redis --stat` - provided `redis` is added to `WARDEN_CMD_ANYARGS` in `bin/warden`

docs/usage.md
```diff
@@ -36,11 +36,14 @@ Flush varnish:
...
-    warden env exec redis redis-cli
+    warden redis
 
 Flush redis completely:
 
-    warden env exec -T redis redis-cli flushall
+    warden redis flushall
+
+Run redis continous stat mode
+    warden redis stat
```
davidalger, line 46: Is it still useful to have this shortcut if `warden redis --stat` functions? Open to either, but we should update `WARDEN_CMD_ANYARGS` (see here: https://github.com/davidalger/warden/blob/develop/bin/warden#L41) to the following so that Warden will pass along the args having a leading double hyphen instead of failing with an unkown arg error:

```
declare WARDEN_CMD_ANYARGS=(svc env db redis sync shell debug)
```

davidalger (commented):
Thanks @werfu and congrats on your first contribution here. Left a couple suggestions below. Great idea on the additional command!

commands/redis.cmd
werfu: You're totally right. I'm not exactly that versed into Bash scripting.

docs/usage.md
ihor-sviziev: I think it's better to use `warden redis --stat` here. What do you think?

davidalger:
Merged and included in the newly tagged 0.12.0 release. Appreciate the contribution!

PR #414, opened 2021-08-20
Add Nginx 1.18 image to wardenenv images

darrakurt:
Just add newest version of nginx server to wardenenv images


davidalger:
Thanks for the contribution @darrakurt. The new docker image should build and be available on Docker Hub shortly.

Issue #220, opened 2020-08-26
WSL2 - Issue with upgrading


ihor-sviziev:
After running `brew upgrade davidalger/warden/warden` (upgrade from 0.9.0 to 0.9.1) and `warden svc up` I got the following message:
![image](https://user-images.githubusercontent.com/1873745/91284165-3646b100-e794-11ea-80ac-273b0447d78d.png)

Everything else works fine

Here is end of the output of `bash -x warden install`
```
+ [[ -f /home/linuxbrew/.linuxbrew/Cellar/warden/0.9.1/commands/install.cmd ]]
+ WARDEN_CMD_VERB=install
+ WARDEN_CMD_EXEC=/home/linuxbrew/.linuxbrew/Cellar/warden/0.9.1/commands/install.cmd
+ WARDEN_CMD_HELP=/home/linuxbrew/.linuxbrew/Cellar/warden/0.9.1/commands/install.help
+ shift
+ ((  0  ))
+ [[ -n '' ]]
+ source /home/linuxbrew/.linuxbrew/Cellar/warden/0.9.1/commands/install.cmd
++ [[ ! -n /home/linuxbrew/.linuxbrew/Cellar/warden/0.9.1 ]]
++ source /home/linuxbrew/.linuxbrew/Cellar/warden/0.9.1/utils/install.sh
+++ [[ ! -n /home/linuxbrew/.linuxbrew/Cellar/warden/0.9.1 ]]
++ [[ ! -d /home/ihor/.warden/ssl/rootca ]]
++ [[ ! -f /home/ihor/.warden/ssl/rootca/private/ca.key.pem ]]
++ [[ ! -f /home/ihor/.warden/ssl/rootca/certs/ca.cert.pem ]]
++ [[ linux-gnu =~ ^linux ]]
++ [[ -d /etc/pki/ca-trust/source/anchors ]]
++ [[ linux-gnu =~ ^linux ]]
++ [[ -d /usr/local/share/ca-certificates ]]
++ [[ ! -f /usr/local/share/ca-certificates/warden-proxy-local-ca.crt ]]
++ [[ linux-gnu == \d\a\r\w\i\n* ]]
++ [[ linux-gnu =~ ^linux ]]
++ [[ ! -f /home/ihor/.warden/nodnsconfig ]]
++ systemctl status NetworkManager
++ grep 'active (running)'
System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to connect to bus: Host is down
[...]

PR #422, opened 2021-09-10
Document Windows support via WSL2

hardyjohnson:
Compilation of Windows installation instructions largely based on my experience and comments in #220. 

@ihor-sviziev I borrowed these images from your comment, hopefully that is OK, otherwise I can re-create.


ihor-sviziev:
@hardyjohnson i don’t mind

davidalger:
Thanks for the contribution @hardyjohnson.

PR #424, opened 2021-09-14
Fix WSL2 paths on installations page

ihor-sviziev:
Follow up on https://github.com/davidalger/warden/pull/422

https://docs.warden.dev/installing.html#windows-installation-via-wsl2
Before:
image
After:
image

This change treats this path as a code to not lose the backslashes


davidalger:
Great catch @ihor-sviziev

Issue #423, opened 2021-09-13
Environment variables are not set when using docker-compose V2


hanhpv:
Today I'm trying to setup Magento 2 with Warden on a Mac, with latest Docker desktop installed. Got these messages when starting env services:

![Clipboard - September 12, 2021 10_28 AM](https://user-images.githubusercontent.com/5069672/133070258-5d989f91-08a4-4fea-ac1c-8951f08c1f8b.png)

It took me few hours to figure out the problem is latest Docker desktop is shipped with Compose V2 RC enabled by default:

![Clipboard - September 13, 2021 3_10 PM](https://user-images.githubusercontent.com/5069672/133070561-d12ef1d2-c249-4338-a9e9-7b5183ac556c.png)

I downgraded docker-compose to version 1.25.0 and it works well.

hanhpv:
This is the setting to downgrade Compose in Docker Desktop.
![Clipboard - September 13, 2021 3_49 PM](https://user-images.githubusercontent.com/5069672/133071262-7008dc30-dc51-427c-9c28-7563a0a91be1.png)

intray:
Amazing, I also solved the same problem when I downgraded docker-compose to version 1.29.2 .

brendanfalkowski:
Just had the same issue on a Mac (just downloaded Docker fresh). Disabling that experimental feature worked.

ihor-sviziev:
This issue isn’t reproducing anymore. More info https://github.com/davidalger/warden/pull/437#issuecomment-1002186894
I think we can close it

PR #437, opened 2021-10-21
Docker compose v2

alinalexandru:
#423 Export variables without default value

I am using this fix, and everything seems to work.


pchoate-trellis:
@alinalexandru Just tested this locally and it works perfectly. Thanks!

pchoate-trellis:
I've been using this change locally for about a week now, but just noticed that the `COMPOSER_VERSION` variable in .env isn't forwarded to the PHP-FPM yml when it is compiled. I'm not entirely certain where/why that is yet.

alinalexandru:
Notised that more variables are not forwarded. I'll investigate

ihor-sviziev:
Looks like this issue isn't reproducing anymore in the `Docker Compose version v2.2.1`

ihor-sviziev:
It worked perfectly fine with the latest docker-compose version for a few weeks.
I think we can close this PR as the issue was fixed on the docker-compose level

davidalger:
Closing, as the issue was a BC break in an upstream dependency.

PR #444, opened 2021-11-04
Increase default client_max_body_size to 5MB

navarr:
1MB is fairly limiting for some very large Magento attributes, and since this is not intended to be used on production, I can't think of any negatives


tdgroot:
@navarr is it possible to override the `client_max_body_size` with a `custom.conf` after this change?

Issue #449, opened 2021-11-12
Feature Request: add tableplus command for mac os


rbouma:
Add a new command to open tableplus automatically for mac users.

rbouma:
any updates on the pr?

rbouma:
@davidalger I removed the install changes for de identifyFile in de ssh config so this is only for the table plus command.

PR #450, opened 2021-11-12
Feature/add tableplus command

rbouma:
PR for feature described in: #449

### DIFF
--- a/commands/tableplus.cmd
+++ b/commands/tableplus.cmd
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+[[ ! ${WARDEN_DIR} ]] && >&2 echo -e "\033[31mThis script is not intended to be run directly!\033[0m" && exit 1
+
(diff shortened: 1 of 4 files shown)


rbouma:
Screenshot 2021-11-12 at 12 13 50

As you can see tableplus is using `~/.ssh/config` for the configuration, edited the install script of warden to prepend the ssh tunnel key to the ssh config of the user only for Mac os users

rbouma:
Leave it open

davidalger:
> As you can see tableplus is using ~/.ssh/config for the configuration

TablePlus uses this, yes, but it also uses the global ssh config file at `/etc/ssh/ssh_config`. The issue here isn't with Warden. It is caused by having `Host *` in the users ssh config, thus overriding the `IdentityFile` configuration.

My recommendation would be to remove the `IdentityFile` directive from the config, as this is what's breaking things here, and it's not really doing much since it's configuring what is already the default value for `IdentityFile`

Issue #388, opened 2021-06-08
Option to disable dnsmasq


davidhiendl:
It would be helpful to have an option to disable dnsmasq svc entirely. Currently it requires stopping/removing the dnsmasq service or removing it from warden's global docker-compose file.

The reason for this is that some people run dnsmasq/unbound/other alternatives directly host with similar rules for other environments, local corporate networks and other dns tools. Running dnsmasq in docker can cause several problems (like #344 but there are more problems), especially with DNS being unavailable before the dnsmasq container is up.

The default should still be enabled, but an option to disable it would be helpful.
If this is acceptable I would be happy to create a PR that disables it if `~/.warden/.env` contains a line like `DNSMASQ_DISABLE=true` or a flag like `~/.warden/.dnsmasq-disabled`

All that is needed for warden to function in such a standalone setup is to add a rule to those services to resolve *.test to 127.0.0.1 which is straight forward for custom configurations.

davidhiendl:
This is still relevant.

davidhiendl:
This is still relevant. PR submitted for review.

PR #462, opened 2021-12-15
Allow dnsmasq service to be disabled 

davidhiendl:
This PR is a solution I've been using in response to my issue #388 "Option to disable dnsmasq".
 
When running a dedicated customized dnsmasq (or similar service) on host or outside the scope of warden it might be in conflict.
This PR adds a configuration flag via `$WARDEN_HOME_DIR/.env` with the key `WARDEN_DNSMASQ_ENABLE` which allows disabling dnsmasq by adding `WARDEN_DNSMASQ_ENABLE=0`.

The default behavior to start a dnsmasq service remains unchanged.

In addition I've added support for loading an additional docker-compose file from ``$WARDEN_HOME_DIR/docker-compose.yml` which is added if present. This allows for deploying a customized dnsmasq (or other services that multiple projects might consume) with easy access to warden docker network and variables.


commands/svc.cmd
```diff
@@ -18,6 +18,28 @@ DOCKER_COMPOSE_ARGS=()
...
+
+## check if dnsmasq is enabled
+if [[ -f "${WARDEN_HOME_DIR}/.env" ]]; then
+    eval "$(grep "^WARDEN_DNSMASQ_ENABLE" "${WARDEN_HOME_DIR}/.env")"
+fi
+
+## add dnsmasq docker-compose
+WARDEN_DNSMASQ_ENABLE="${WARDEN_DNSMASQ_ENABLE:-1}"
+if [[ "$WARDEN_DNSMASQ_ENABLE" == "1" ]]; then
+    echo "adding docker-compose.dnsmasq.yml"
```
ihor-sviziev, line 30: I think this is useless message
```suggestion
```

ihor-sviziev (changes requested):

ihor-sviziev (approved):

commands/svc.cmd
davidhiendl: Echo statement removed. Sorry for taking so long, this somehow slipped my mind.

davidhiendl:
I'm not exactly sure why these tests failed, my changes do not relate to them at all and the logs seem to point to a node-sass compiler issue. What can I do to fix this?

commands/svc.cmd
```diff
@@ -18,6 +18,27 @@ DOCKER_COMPOSE_ARGS=()
 DOCKER_COMPOSE_ARGS+=("-f")
 DOCKER_COMPOSE_ARGS+=("${WARDEN_DIR}/docker/docker-compose.yml")
 
+
+## check if dnsmasq is enabled
+if [[ -f "${WARDEN_HOME_DIR}/.env" ]]; then
+    eval "$(grep "^WARDEN_DNSMASQ_ENABLE" "${WARDEN_HOME_DIR}/.env")"
+fi
```
Den4ik, line 25: This part is not required due to `docker-compose up` extract variables from `.env` file automatically

Den4ik (changes requested):
Hi @davidhiendl 
Thanks for this PR. Please look at my comment

commands/svc.cmd
davidhiendl: @Den4ik Actually I think it is required because the variable is not needed in `docker-compose up` but in the bash script itself to determine IF the additional docker-compose file is added to the arguments list. I just tried it without that check and `set -x` and it is not defined at that point resulting in dnsmasq starting regardless of the value set in `.env`.

Den4ik: Yep, my mistake, you where right

Den4ik (approved):

davidhiendl:
@Den4ik How can I re-run the tests? I believe this was a temporary failure.

navarr (approved):

PR #469, opened 2022-01-10
Updated environment default Elasticsearch version from 7.6 to 7.13

centerax:
7.6 does not run on M1 and 7.13 does.

### DIFF
--- a/docs/environments/magento2.md
+++ b/docs/environments/magento2.md
@@ -32,7 +32,7 @@ The below example demonstrates the from-scratch setup of the Magento 2 applicati
 
         WARDEN_SYNC_IGNORE=
 
-        ELASTICSEARCH_VERSION=7.6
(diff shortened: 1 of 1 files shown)


ihor-sviziev:
Hi @centerax,
The issue is not related to the elasticsearch version but because the image is built for x86_64 architecture. I created PR for re-building the images for ARM64 in https://github.com/davidalger/warden/pull/488

davidalger:
Not opposed to changing the default versions used for new environments, but it should follow the system requirements of the target platform (in this case, Magento 2): https://devdocs.magento.com/guides/v2.4/install-gde/system-requirements.html

ES 7.13 isn't mentioned, although highly unlikely to cause an issue IMO, but if we're going to update the default here, let's bump it to ES 7.17, which is both the latest available and the version officially supported by Magento 2.4.5

PR #472, opened 2022-01-18
Small docs improvement

tjadowski:
Use of active firewall on host needs a new firewall rule to allow xdebug connection.


davidalger:
Thanks @tjadowski

PR #480, opened 2022-02-15
Remove varnish 4.1 support

ihor-sviziev:
My PR removes building of varnish 4.1 image as it's currently built incorrectly - it contains varnish 6.0:

Screenshot 2022-02-15 at 09 26 24

https://github.com/davidalger/warden/runs/5027750141?check_suite_focus=true#step:3:921


ihor-sviziev:
Ohh... Looks like all the images are built with Varnish 6.0... 
https://github.com/davidalger/warden/runs/5027750141?check_suite_focus=true#step:3:2046

davidalger:
@ihor-sviziev Thanks for bringing this to light. Seems something (I have no idea what) caused the appstream module to override the rpm from the package cloud repo. This should be fixed by #482 as soon as the images build.

PR #483, opened 2022-02-15
Move varnish images to almalinux

Den4ik:
This PR replace CentOS images with AlmaLinux https://almalinux.org/


ihor-sviziev:
Since we moved in https://github.com/davidalger/warden/commit/e384f5665527e88ca7880a6ea067b3cc9a2829b6 to centos8 stream that has EOL on May 31st, 2024, I think we're fine to keep it as is.

PR #484, opened 2022-02-15
Varnish multi-arch builds

ihor-sviziev:
This PR adds ARM64 architecture builds for Varnish images

**Status:**
-  ❌ Varnish 4.1 image isn't supporting arm64 architecture as it's based on centos 7 that don't have arm64 support
- ✅ Varnish 6.0
- ❌ Varnish 6.4 image for which [we don't have arm64 packages](https://packagecloud.io/varnishcache/varnish64)
- ✅  Varnish 6.5 image
- ✅  Varnish 6.6 image (added in this PR)
- ✅  Varnish 7.0 image (added in this PR)

All images are currently available at  https://github.com/ihor-sviziev/warden/pkgs/container/varnish, they're built from the following commit: https://github.com/ihor-sviziev/warden/commit/727e086d3731f4cbda557ce3481d43a0d5b93f3d

Note: Marked with ❌ symbol images are old images and they won't support arm any time soon.

### DIFF
--- a/.github/workflows/docker-image-varnish.yml
+++ b/.github/workflows/docker-image-varnish.yml
@@ -1,18 +1,18 @@
 name: Docker Image Varnish
 on:
   schedule:
-  - cron: "0 6 1 * *"   # 6 AM UTC on 1st day of month
+    - cron: "0 6 1 * *"   # 6 AM UTC on 1st day of month
   push:
     paths:
-    - version
-    - images/.trigger
-    - images/varnish/**
+      - version
+      - images/.trigger
+      - images/varnish/**
     branches:
-    - develop
+      - develop
 jobs:
   varnish:
     name: Varnish
-    
+
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


ihor-sviziev:
About the 4.1 image - it uses x86_64 base image, so it can't support arm64 architecture.

About the 6.4 image - just looked - the arm packages are simply not available, but available for [6.5](https://packagecloud.io/varnishcache/varnish65), [6.6](https://packagecloud.io/varnishcache/varnish66), [7.0](https://packagecloud.io/varnishcache/varnish70)

Maybe we should build the multi-arch images for 6.0, 6.5, and add 6.6, 7.0 release lines. What do you think?

ihor-sviziev:
Okay, so now I built successfully the arm builds for varnish 6.0, 6.5, 6.6 and 7.0.
All the images are available at https://github.com/ihor-sviziev/warden/pkgs/container/varnish. They were built with https://github.com/ihor-sviziev/warden/runs/5208800122?check_suite_focus=true

Will revert all not related changes

ihor-sviziev:
@davidalger, could you please review this PR?

convenient:
@ihor-sviziev You can use a custom image to make Varnish 4.1 work?


```
# must be a multi arch image that supports amd64 and arm64
# https://hub.docker.com/_/debian?tab=tags&page=1&name=stretch-20211115-slim
FROM debian:stretch-20211115-slim

# https://varnish-cache.org/docs/4.1/installation/install.html#compiling-varnish-from-source
RUN apt-get update && \
    apt-get install -y pkg-config automake tree autotools-dev libtool build-essential curl git libedit-dev libjemalloc-dev libncurses-dev libpcre3-dev python3-sphinx python3-docutils

RUN git clone --single-branch --branch varnish-4.1.11 https://github.com/varnishcache/varnish-cache/
WORKDIR varnish-cache
RUN ./autogen.sh && \
    ./configure && \
    make && \
    make install && \
    ldconfig && \
    varnishd -V 2>&1 | grep "varnish-4.1.11"

EXPOSE 80

COPY default.vcl /etc/varnish/default.vcl

# The sleep is here because it takes a second for `varnishd` to create the
# `_.vsm` file that `varnishlog` reads (on a local VM at least)
CMD varnishd -p vcl_dir=/etc/varnish/ -a :80 -f /etc/varnish/default.vcl -p http_max_hdr=128 && \
    sleep 1 && \
    varnishlog

```

ihor-sviziev:
I think we shouldn't rebuild and add support for old and reached the end of life versions.
https://varnish-cache.org/releases/



Release | Date | EOL Date | Download
-- | -- | -- | --
7.0.2 | 2022-01-25 | 2022-09-15 | varnish-7.0.2.tgz
6.6.2 | 2022-01-25 | 2022-03-15 | varnish-6.6.2.tgz
6.0.10 | 2022-01-25 | Supported | varnish-6.0.10.tgz

> All releases not mentioned above are End-Of-Life and unsupported.

convenient:
Agreed its EOL was just showing the approach that would maintain current feature parity (assuming 4.1 current works?)

ihor-sviziev:
@davidalger cool! I also added few other commits in my own branch. Feel free to copy the changes from https://github.com/ihor-sviziev/warden to make it available for everyone

PR #487, opened 2022-02-17
Fix homepage logo

ihor-sviziev:
Add missing spaces at the end of the logo to fix the logo

**Before:**
image

**After:**
image


davidalger:
Very nice catch!

PR #488, opened 2022-02-18
Elasticsearch multi-arch builds

ihor-sviziev:
This PR adds ARM64 architecture builds for Elasticsearch images
**Status:**
❌ Elasticsearch 6.8 doesn't support arm builds as base image is not optimized.
✅ Elasticsearch 7.13 does support arm builds

All images are currently available at https://github.com/ihor-sviziev/warden/pkgs/container/elasticsearch

Note: Marked with ❌ symbol images are old images and they won't support arm any time soon.

### DIFF
--- a/.github/workflows/docker-image-elasticsearch.yml
+++ b/.github/workflows/docker-image-elasticsearch.yml
@@ -12,11 +12,47 @@ on:
 jobs:
   elasticsearch:
     name: Elasticsearch
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - version: '6.8'
+            multiarch: 'false'
+          - version: '7.13'
+            multiarch: 'true'
+
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v1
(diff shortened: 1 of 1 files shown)


convenient:
>❌ Elasticsearch 6.8 doesn't support arm builds as base image is not optimized.


@ihor-sviziev If the base image supports multi arch then the image you're building should support multi arch. So mainline ES don't support it but there are other options.

https://github.com/blacktop/docker-elasticsearch-alpine/issues/14#issuecomment-1047570159

I got this slim version of ES 6.5 running recently.

ihor-sviziev:
@convenient, I'm not sure if we need to support all the listed versions. As to me - we should support the only officially available options and not introduce our custom builds.

convenient:
@ihor-sviziev I mean that's fair on the face of it. I was just saying as it currently supported and may not be necessary to drop it.

Also it's not like there's versions of Magento 2.3.4 out there in the wild that will one day need onboarded and upgraded 🤷  But totally agree can't support every hypothetical deprecated version out there.

PR #489, opened 2022-02-21
Require docker compose 2.2.3 when compose v2 is active

### LINKED ISSUE #423: Environment variables are not set when using docker-compose V2

davidalger:
As noted in #423 compose v2 has seen some dramatic issues. This resolves #423 by enforcing a constraint of 2.2.3 or higher while still allowing continued use of v1 with existing constraint of 1.25.0 or greater, as there may still be compatibility issues due to upstream BC breaks and Linux users will still have v1 installed by default (v2 is now the default for Docker Desktop on macOS)

### DIFF
--- a/bin/warden
+++ b/bin/warden
@@ -22,9 +22,16 @@ fi
 
 ## verify docker-compose meets version constraint
 DOCKER_COMPOSE_REQUIRE="1.25.0"
+DOCKER_COMPOSE_REQUIRE_V2="2.2.3"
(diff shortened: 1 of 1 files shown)


davidalger:
Adding arm64 label as compose v2 is required for fully native experience on Apple Silicon.

PR #493, opened 2022-02-22
Feature/add tableplus command

rbouma:
PR for feature described in: https://github.com/davidalger/warden/issues/494

### DIFF
--- a/commands/tableplus.cmd
+++ b/commands/tableplus.cmd
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+[[ ! ${WARDEN_DIR} ]] && >&2 echo -e "\033[31mThis script is not intended to be run directly!\033[0m" && exit 1
+
+if [[ "$OSTYPE" != "darwin"* ]]; then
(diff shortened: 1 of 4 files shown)


rbouma:
@davidalger I removed the install changes for de identifyFile in de ssh config so this is only for the table plus command.

ihor-sviziev (changes requested):

tdgroot:
Not a big fan of this feature. Not everyone is using tableplus, it's a non-free program and it feels like feature creep to me. I'd rather have detailed guides in the docs on how to configure a database editor.

davidalger:
Tend to agree with @tdgroot's reasoning on this one. Personally a huge fan of Table Plus, but integrations and shortcuts with every possible database client isn't something this project can reasonably be expected to have. What I would recommend is putting your work here into your project's `.warden/commands` directory, or placing it at `~/.warden/commands/tableplus.cmd` where Warden should load it (https://github.com/davidalger/warden/blob/develop/bin/warden#L50-L77) and be able to execute it from anywhere.

PR #496, opened 2022-03-08
Ignore the only root .git directory

ihor-sviziev:
While working with m2 projects, it's quite a frequent case when you have a separate git repository per extension, so while developing something you're installing a module from a branch, do your changes directly in vendor/<vendor_name>/<module_name> and then do commits directly from this. Such flow helps significantly reduce time on installing each new version via composer.

The issue with Warden in such configuration is that ALL .git folders are ignored in the mutagen config. As a result - on Mac OS, you can't see the diff, commit changes, etc. (the .git directory is entirely missing in the module folder) 

My PR changes this behavior by ignoring the only root directory git repo.

### DIFF
--- a/environments/magento2/magento2.mutagen.yml
+++ b/environments/magento2/magento2.mutagen.yml
@@ -5,8 +5,11 @@ sync:
     watch:
       pollingInterval: 10
     ignore:
-      vcs: true
+      vcs: false
       paths:
+        # Root .git folder 
+        - "/.git/"
+      
         # System files
         - ".DS_Store"
         - "._*"


dudzio12:
I temporary use:

```
WARDEN_SYNC_IGNORE=!vendor/vendor-name/*/.git
```

inside `.env` config file.

Den4ik:
I faced with pure performance on one of my projects when ignored vsc.
So currently I use version control only from MacOS and don't use it inside containers

ihor-sviziev:
@navarr, what do you think about porting this change to https://github.com/swiftotter/den?

navarr:
@ihor-sviziev Why have the git repositories in the container instead of on the host machine?

ihor-sviziev:
@navarr as I described in the original message, the use case when you're having a single module in few magento 2 projects. Usually to simplify development we're installing the module via `composer update <vendor_name>/<module_name> --prefer-source`, that installs module as a git repository. Then we just add needed changes in that module's directory (`/vendor/<vendor_name>/<module_name>/`) and run git add, git commit, git push from the module's directory.

Right now, mutagen sync excludes all the git repositories from the sync, so it mean inside container we do have .git directory, but on the host machine we don't, so we can't commit and push from the host machine.

ihor-sviziev:
@navarr what do you think about porting this change to den?

navarr:
@ihor-sviziev I think this is likely a good and useful change.  Can you open a PR against Den?

PR #501, opened 2022-03-28
Bring in changes from drpayyne's fork

navarr:
I'm unsure why a PR doesn't already exist, but thought I'd open this one.

I've been using this as my daily driver for months now and so far have no concerns.

What needs doing/testing to make this update happen?


drpayyne:
Indeed, myself and so many more developers whom I just know in the agency I work at, use my fork as a daily driver for all our projects. So it's definitely working and stable! :)

davidalger:
Many of the images have been addressed in #505. Still a few (php!) pending. Closing since merging the fork in without changes isn't feasible, so going to close this, but I don't plan to simply let this drop

schizek:
@davidalger, Is there any update on the status of supporting Mac M1s?

drpayyne's fork is the only way my team can use Warden with Magento 2 on my Mac M1 machines.  On the current implementation of Warden, Magento 2 runs really slow, Mailhog doesn't work for me at all and Elasticsearch is spotty only working some of the time.  However, the fork does not work for WordPress implementations so definitely some work needs to be done there to get it fully functional with more than just Magento.

PR #512, opened 2022-04-11
Fixing unclear docs for configuring magento 2 credentials

Nuuttif:
https://github.com/davidalger/warden/issues/511


Nuuttif:
Why do these checks fail? I just changed a few lines from docs and I don't think these changes relate to any errors.

davidalger:
@Nuuttif Sorry about the failing checks, that's wasn't your issue; fixed that this morning in #514 so the MR here could build and setup the deploy preview for me.

Made one tweak and merged the fix. Thanks for the callout and MR to resolve it.

PR #526, opened 2022-05-05
Only use stale bot for issues

sprankhub:


sprankhub:
:joy: That would be funny if it wasn't so sad...

PR #542, opened 2022-09-13
Add DBeaver docs

bbutkovic:
This adds documentation for connecting to DBs through DBeaver.


navarr:
Docs are moving to wardenenv/docs

PR #582, opened 2023-01-20
Ported over markshust commands

rbouma:
I have ported all Markshust stack commands for easy transfer from markshust to warden.
And for easy use in new commands.


navarr:
I like this, but I think a lot of these are very Magento-specific.  I'm hesitant to load them all into the global namespace.

bap14:
I agree with @navarr . This would be an interesting case for having environment specific commands that are added in / available when called. So similar to the project-specific commands in `{PROJECT_DIR}/.warden/commands`, might be worth adding a `commands` directory to the environment and having Warden set the preference order to be (from most preferred to least preferred):

# Project-Specific Command (`${WARDEN_ENV_DIR}/.warden/commands/*.cmd`)
# Environment-Specific Command (`${WARDEN_DIR}/environments/${WARDEN_ENV_TYPE}/commands/*.cmd`)
# Personal Global Command (`${WARDEN_HOME_DIR}/commands/*.cmd`)
# Global Command (`${WARDEN_DIR}/commands/*.cmd`)

This may be a breaking change, as at the moment I believe the preference is (from most preferred to least preferred):

# Global Command (`${WARDEN_DIR}/commands/*.cmd`)
# Project-Specific Command (`${WARDEN_ENV_DIR}/.warden/commands}/*.cmd`)
# Personal Global Command (`${WARDEN_HOME_DIR}/commands/*.cmd`)

PR #584, opened 2023-01-20
Added vuejs environment

rbouma:
Added Vuejs environment + nginx config in https://github.com/wardenenv/images/pull/2


navarr:
This looks like a simple static not-even-PHP environment. I think a different approach for this that's not vue-specific (since no part of it seems to be) would be in order.

I've always thought Warden should have a generic `php` environment, and perhaps a generic `static` environment would be useful too (that does nothing but stand-up Nginx and mount into it)

rbouma:
@navarr Agree added simple php environment https://github.com/wardenenv/warden/pull/590 will add the static later.

PR #590, opened 2023-01-27
Added simple php environment type

rbouma:


navarr (changes requested):
`ROLL_` means this isn't ready for Warden proper

PR #623, opened 2023-03-21
Add Backup and Restore commands

navarr:


commands/backup.cmd
```diff
@@ -0,0 +1,107 @@
...
+            $BACKUP_LOC:/backup ubuntu bash \
+            -c "tar -czvf /backup/redis.tar.gz /data"
+
+    ;;
+    elasticserach)
+
+        docker run \
+            --rm --name $CONTAINER_NAME \
+            --mount source=$ES_VOLUME,target=/data -v \
+            $BACKUP_LOC:/backup ubuntu bash \
```
rootindex, line 64: Same suggestion as made earlier.

commands/backup.cmd
```diff
@@ -0,0 +1,107 @@
...
+    redis)
+
+        docker run \
+            --rm --name $CONTAINER_NAME \
+            --mount source=$REDIS_VOLUME,target=/data -v \
+            $BACKUP_LOC:/backup ubuntu bash \
+            -c "tar -czvf /backup/redis.tar.gz /data"
+
+    ;;
+    elasticserach)
```
bap14, line 59: Typo. This should be "elasticsearch"

bap14:
While I understand the point of this PR, I do wonder if this is the right way to approach a "backup" solution. I'm not sure about redis and elasticsearch, but I've seen times where using the data of one MariaDB version with another prevents MariaDB from starting.

It's safer, and more portable, to just generate an SQL dump into a file (or files). I'm not sure if redis or Elasticsearch have the same mechanism (redis probably not, but ES maybe).

commands/backup.cmd
bap14, line 59: You should look to add support for OpenSearch here as well, since Warden offers the choice between ES and OS. I'm not certain if the data location is the same (can't see why it would be different) but just something to look at adding.

bap14:
Another thing to consider is that this is a generically named "backup" command; however, it's very clearly tailored to a Magento install. If you run the "all" command it's assuming that all the services (DB, Redis, ES) will be utilized in the project; however, not all environments need/use them (e.g. Magento 1, Drupal, etc.).

PR #630, opened 2023-03-22
Fix db.base.yaml from Mariadb

israelguido:
Error when pulling Mariadb, the tag was fixed.
Screen Shot 2023-03-22 at 18 06 36


navarr:
We've identified that this is an issue in Docker Compose v1. As it is deprecated and we plan to convert commands to `docker compose` this fix will not be accepted (and is not acceptable as it stands as it doesn't allow DB_DISTRIBUTION_VERSION)

In the interim you can update your docker compose standalone to v2

Issue #527, opened 2022-05-08
Not compatible with composer V2 on Linux


guvra:
Hi,

I installed Docker compose v2 with the [official APT repository](https://docs.docker.com/engine/install/ubuntu/#install-using-the-repository):

```
curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo gpg --dearmor -o /usr/share/keyrings/docker-archive-keyring.gpg
echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/docker-archive-keyring.gpg] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable" | sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
sudo apt-get update
sudo apt-get install docker-ce docker-ce-cli containerd.io docker-compose-plugin
```

With the above setup, docker compose is only available as a plugin:

```
$ docker compose version
Docker Compose version v2.4.1

$ docker-compose
Command 'docker-compose' not found
```

Unfortunately, Warden fails to detect compose:

```
$ warden
ERROR: docker-compose could not be found; please install and try again.
```

Adding an alias in ~/.bashrc does not resolve the issue.

Adding a custom docker-compose script in /usr/local/bin solves the detection issue, but warden still triggers an error:

```
$ warden
ERROR: docker-compose version should be 1.25.0 or higher (. installed)
```

In the `warden` script, the version is checked with `docker-compose --version`, but the command fails on compose v2. The version can only be displayed with `docker compose version`.

ihor-sviziev:
I guess the solution will be replacing all `docker-compose` commands to `docker compose`, but that increases the minimum docker-compose version to 2.3+

springerin:
The docker-compose version 1.x can be installed along aside the new version. Obviously is not using the new version but till warden finds a way to be compatible with both one can still work with the old one for warden while using the new version for everything else.
One other way could be the use of https://github.com/docker/compose-switch but I didn't test it myself so can't be sure how good of a solution it is.

kbrownlow:
This is what I did to solve:

Open Docker Desktop.
Go to Settings > General.
Tick the "Enable Docker Compose V1/V2 compatibility mode" setting.
Click "Apply & Restart".

That got it working for me instantly.

guvra:
Bad bot.

Issue #625, opened 2023-03-21
Migrate all `docker-compose` commands to `docker compose` commands


navarr:
Will resolve #527, #622

sprankhub:
Since quite some users seem to bump into this, I think this should have high priority :see_no_evil: Also, according to https://docs.docker.com/compose/compose-v2/#differences-between-compose-v1-and-compose-v2, it could be as easy as replacing `docker-compose` to `docker compose` without any additional changes:

> The `compose` command in the Docker CLI supports most of the `docker-compose` commands and flags. It is expected to be a drop-in replacement for `docker-compose`.

Do you see any specific things to watch out for, @navarr? If not and you do not get to it, I'd consider sending a PR.

navarr:
@sprankhub I thought there were.  I previously did this in Den but I only see the one commit, so I've ported that over in #631 

If you have the capability to test this particular issue, I'd appreciate some additional eyes on it.

sprankhub:
@navarr, sorry for the late feedback. I've installed the `docker-compose-plugin` branch from the PR to `/opt/warden` and ran `/opt/warden/bin/warden env pull && /opt/warden/bin/warden env up --force-recreate`. From the fancier output, I can tell that indeed, docker compose v2 is used. The project still seems to run fine, so I think we are good. Could you merge and release a new version soonish? Looks like there were quite some people affected by this. Thanks!

navarr:
@sprankhub unfortunately that's docker compose v1 vs v2. I can do some research to see if we're stuck with it or not but so far I've always presumed so

sprankhub:
Thanks for the info, @navarr. Given that it is a one-time change, I'm fine with it. More important to make the switch to docker compose v2 :)

Issue #622, opened 2023-03-20
[Image] invalid tag format for MARIADB_VERSION=10.6


sprankhub:
### Version of Warden

0.13.0

### Operating System and Installation Method

Ubuntu 22.04 via Brew

### Image

db / MARIADB_VERSION=10.6

### Describe the Bug

I still have the "old" DB version definition in my `.env` file:

    MARIADB_VERSION=10.6

I get an error when updating the images:

    $  warden env pull && warden env up
    Pulling db               ... error
    Pulling redis            ... done
    Pulling rabbitmq         ... done
    Pulling elasticsearch    ... done
    Pulling php-fpm          ... done
    Pulling nginx            ... done
    Pulling varnish          ... done
    Pulling php-debug        ... done
    Pulling elasticsearch-hq ... done

    ERROR: for db  invalid tag format
    ERROR: invalid tag format

### To Reproduce

1. Add `MARIADB_VERSION=10.6` to your `.env` file.
2. Run `warden env pull && warden env up`.
3. See "error" on "Pulling db". Error: invalid tag format

navarr:
Hi @sprankhub, could you please supply the output of `warden env config`?  Feel free to remove any sensitive information from it if applicable

sprankhub:
networks:
      default:
        labels:
          dev.warden.environment.name: magento
          dev.warden.environment.type: magento2
    services:
      db:
        command:
        - mysqld
        - --max_allowed_packet=1024M
        - --explicit_defaults_for_timestamp=on
        environment:
          MYSQL_DATABASE: magento
          MYSQL_HISTFILE: /sql_history/.sql_history
          MYSQL_PASSWORD: magento
          MYSQL_ROOT_PASSWORD: magento
          MYSQL_USER: magento
        hostname: magento-mariadb
        image: docker.io/wardenenv/mariadb:$${MARIADB_VERSION:-10.4}
        volumes:
        - sqlhistory:/sql_history:rw
        - dbdata:/var/lib/mysql:rw
      elasticsearch:
        environment:
          ES_JAVA_OPTS: -Xms64m -Xmx512m
          discovery.type: single-node
          xpack.security.enabled: "false"
        hostname: magento-elasticsearch
        image: docker.io/wardenenv/elasticsearch:7.17
        labels:
          traefik.docker.network: magento_default
          traefik.enable: "true"
          traefik.http.routers.magento-elasticsearch.rule: Host(`elasticsearch.magento.test`)
[...]

PR #631, opened 2023-03-23
Move from docker-compose standalone to `docker compose` commands

navarr:
Issues #625, #622, #527


davidg-klevu:
For what it's worth, this change resolved #527 / #622 issues for me :slightly_smiling_face:

PR #665, opened 2023-06-19
Move from MYSQL_DISTRIBUTION -> DB_DISTRIBUTION

navarr:
Part of ensuring bc-compatibility with Den


environments/magento1/init.env
```diff
@@ -2,8 +2,8 @@
 WARDEN_DB=1
 WARDEN_REDIS=1
 
-MYSQL_DISTRIBUTION=mariadb
-MYSQL_DISTRIBUTION_VERSION=10.3
+DB=mariadb
```
bap14, line 5: @navarr Is this missing the `_DISTRIBUTION` part of the key or is this a weird edge case?

navarr: Good catch.  Appreciate you reviewing my WIPs.  <3

environments/magento1/init.env
```diff
@@ -2,8 +2,8 @@
...
+DB_VERSION=10.3
```
navarr, line 6: `
```suggestion
DB_DISTRIBUTION=mariadb
DB_DISTRIBUTION_VERSION=10.3
```

commands/env.cmd
```diff
@@ -9,6 +9,9 @@ if (( ${#WARDEN_PARAMS[@]} == 0 )) || [[ "${WARDEN_PARAMS[0]}" == "help" ]]; the
   $WARDEN_BIN env --help || exit $? && exit $?
 fi
 
+DB_DISTRIBUTION=${DB_DISTRIBUTION:-${MYSQL_DISTRIBUTION:-"mariadb"}}
+DB_DISTRIBUTION_VERSION=${DB_DISTRIBUTION_VERSION:-${MYSQL_DISTRIBUTION_VERSION:-${MARIADB_VERSION:-"10.4"}}}
```
ihor-sviziev, line 13: 1. shoudln't we add `export` there?
```suggestion
export DB_DISTRIBUTION=${DB_DISTRIBUTION:-${MYSQL_DISTRIBUTION:-"mariadb"}}
export DB_DISTRIBUTION_VERSION=${DB_DISTRIBUTION_VERSION:-${MYSQL_DISTRIBUTION_VERSION:-${MARIADB_VERSION:-"10.4"}}}
```

2. maybe it's better to pring some warning if the DB_DISTRIBUTION and DB_DISTRIBUTION_VERSION variable aren't declared yet? In this case we might suggest changing MYSQL_DISTRIBUTION variable to DB_DISTRIBUTION and MYSQL_DISTRIBUTION_VERSION to DB_DISTRIBUTION_VERSION. What do you think?

ihor-sviziev (changes requested):

commands/env.cmd
navarr: I don't exactly follow #2.

Warden's old default is MariaDB 10.4.  A few versions ago I mistakenly introduced the ability to change database distribution and version using MYSQL_ variables (the mistake)

Moving to DB_DISTRIBUTION so it matches Den is what I'm trying to do here, while maintaining backwards compatibility with any `.env` file that uses MYSQL_ variables, and maintaining backwards compatibility with the original defaults

ihor-sviziev (approved):

ihor-sviziev:
@navarr just wonder, why you closed it?

navarr:
@ihor-sviziev it doesn't feel like there's a point in the backwards compatibility after all this time

PR #666, opened 2023-06-19
Porting in features from Den

navarr:


commands/status.help
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+[[ ! ${WARDEN_DIR} ]] && >&2 echo -e "\033[31mThis script is not intended to be run directly!\033[0m" && exit 1
+
+WARDEN_USAGE="Provides listing of projects that are currently running and Den's Traefik is connected to"
```
bap14, line 4: @navarr Would be worth updating the project name in this string to be "Warden" instead of "Den".

PR #706, opened 2023-09-23
#115: linux permissions

toonvd:
@navarr this is mainly a kickoff

To cleanly set a gid and uid, this should happen on build time.
Since images are build in advance, this is not possible.
Adding an entrypoint in docker-compose was not clean enough either in my opinion, hence checking in the project for php-fpm and php-debug, fixing the user and restarting the container.

### DIFF
--- a/commands/env.cmd
+++ b/commands/env.cmd
@@ -4,8 +4,8 @@
 WARDEN_ENV_PATH="$(locateEnvPath)" || exit $?
 loadEnvConfig "${WARDEN_ENV_PATH}" || exit $?
 assertDockerRunning
-HOSTUID=$(id -u)
-HOSTGID=$(id -g)
+HOST_UID=$(id -u)
+HOST_GID=$(id -g)
 
 if (( ${#WARDEN_PARAMS[@]} == 0 )) || [[ "${WARDEN_PARAMS[0]}" == "help" ]]; then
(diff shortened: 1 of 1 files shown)


ihor-sviziev:
Don't you think we'll have some other permission issues?

toonvd:
@ihor-sviziev as stated, this was a kickoff, if you can point me towards those other permission issues, I can help.

navarr:
(Leaving a comment to let you know this isn't being ignored, I've got my eyes on it I'm just also incredibly busy and will get to it in course 👍)

ihor-sviziev (changes requested):
Hi @toonvd,
I just looked on your changes - looks like they should be moved inside the entrypoint script, similar to this:
https://github.com/wardenenv/images/blob/main/php-fpm/context/docker-entrypoint#L10-L12
Also, we should pass the host uid and host gid to the needed containers.
What do you think about it?

toonvd:
Hi Ihor!

Will do! Closing this one, will reopen one in images soon.
If you know anyone that can help with testing (switch to Mac for my own laptop), please let me know!

ihor-sviziev:
The only thing I can suggest - run a virtual machine :) 
I actually not sure if it's doable in that entrypoint file, but I think that would be the best choise

PR #712, opened 2023-10-16
Use latest Magepack version by default

sprankhub:
Warden by default uses a very old Magepack version: https://github.com/wardenenv/warden/blob/9548566e981c0a4fb61c23f5ca6cab2e48066305/environments/magento2/magento2.magepack.base.yml#L5

We should at least propose using the newest version in the example `.env` file (if not bumping the default version).


ihor-sviziev (approved):

navarr (approved):

ihor-sviziev:
@sprankhub don't you think that the default value should be also updated?

sprankhub:
@ihor-sviziev, I personally think so (I even suggested it above), but this would be kind of a breaking change :thinking:

navarr:
@ihor-sviziev @sprankhub It's a difficult question, which is why I've erred on the side of leaving it alone (for now).

I think at this point, nobody using Warden is expecting an undefined Magepack version to be the latest version.  Such a change would really only benefit new users, while possibly being an unexpected change for older users.

PR #746, opened 2024-02-07
Updated ES_JAVA_OPTS to prevent 429 while indexing Magento

lewisvoncken:


iriks-it:
👍🏻

navarr:
I don't have enough knowledge in this area to know if this is a good change to merge globally.  I'd love more opinions, especially from those more knowledgeable with Java

bap14:
@navarr Xms is the "starting" memory allocated to the JVM, while Xmx is the "maximum" memory that can be allocated to the JVM. So with the current commit they're allocating 2GB at the start with a max of 2GB to consume (meaning it won't grow).

The better approach would be to leave Xms low and increase the Xmx value to something like 1g.

As this seems project-specific, I would question whether bumping up the max memory is advised considering the Docker VM resources on Mac and Windows can be constrained to as little as 1GB. Allowing the JVM to use up to 2GB would have the potential to still run out of memory even if the Xmx value is raised. This feels like another case of adjusting the environment variable for that specific service in the project's `.warden/warden-env.yml` instead of making a global change.

navarr (changes requested):
Based on discussion on this PR, if you're willing to proceed with contributing - my suggestion would be to implement a `WARDEN_SEARCH_MAXMEM` variable that defaults to 512m but can be easily extended on a per-project basis without modifying docker yaml file.

PR #747, opened 2024-02-07
Remove Mutagen because file mount combined with Orbstack is fast!

lewisvoncken:


bap14:
@lewisvoncken This is the wrong approach to take. Warden is not locked to OrbStack as the underlying container platform. It's only locked to `docker` and `docker compose`, be that Docker Desktop, OrbStack, Containerd, Lima, Colima, etc.

I believe the proper way to handle this is to make Mutagen an optional service that can be turned on or off per environment, and potentially globally, using the `.env` file like we can do with ElasticSearch, Varnish or Blackfire. This way we're still backwards compatible with all of the existing Warden releases while allowing those using different Docker platforms to choose to use bind mounts or Mutagen.

Cc. @navarr

joseluisi4:
@bap14, In my experience, Mutagen is no longer necessary in Docker Desktop, in fact, it is counterproductive, I have been using Warden without Mutagen on my Mac m2 for several months and it works much better, I no longer have file desynchronization problems and a notable decrease in memory consumption.
Since Docker Desktop made **VirtioFS** as the default file sharing implementation, Mutagen is not needed.

On the other hand, I think that the `.env` file is not the best place to add the Mutagen configuration since this is a preference of your machine and not of the project.

bap14:
@joseluisi4 I'm not against deprecating Mutagen; however, we have to keep legacy projects and environments in mind. VirtioFS is still not the default in Docker Desktop and so may not be enabled in every environment. If we go down this path, which I'm not against, we should mark it as deprecated in the next release and target the next, or one after next, minor release.

So for example the current release is 0.14.2. In 0.14.3 we implement the ability to disable Mutagen via the `.env` config. We'll also need to add a notice every time Mutagen sync starts via the `warden env up` command to inform that Mutagen is being deprecated. The default is still to use Mutagen. In 0.15 we can swap the default of using Mutagen to default to off while still having the ability to enable it if wanted.

navarr:
Closing this as work proceeds in the replacement PR.

PR #749, opened 2024-02-15
Optional Mutagen

joseluisi4:
This change allows you to activate or deactivate **Mutagen** from the warden configuration `.env` file with `WARDEN_MUTAGEN_ENABLE=0`.
By default, Mutagen is active.
A message is also painted indicating that Mutagen is being deprecated.

### DIFF
--- a/commands/env.cmd
+++ b/commands/env.cmd
@@ -19,6 +19,11 @@ trap '' ERR
 ## define source repository
 if [[ -f "${WARDEN_HOME_DIR}/.env" ]]; then
   eval "$(sed 's/\r$//g' < "${WARDEN_HOME_DIR}/.env" | grep "^WARDEN_")"
+
+  ## Temporary fix to force mutagen activation
+  if [[ ${WARDEN_MUTAGEN_ENABLE} == '' ]]; then
+    export WARDEN_MUTAGEN_ENABLE=1
+  fi
 fi
 export WARDEN_IMAGE_REPOSITORY="${WARDEN_IMAGE_REPOSITORY:-"docker.io/wardenenv"}"
 
@@ -174,7 +179,10 @@ TRAEFIK_ADDRESS="$(docker container inspect traefik \
 )"
 export TRAEFIK_ADDRESS;
 
-if [[ $OSTYPE =~ ^darwin ]]; then
+if [[ $OSTYPE =~ ^darwin ]] && [[ ${WARDEN_MUTAGEN_ENABLE} -eq 1 ]]; then
+    echo -e "\033[31mMutagen is being deprecated!!\033[0m"
+    echo -e "\033[31mTo disable it, add WARDEN_MUTAGEN_ENABLE=0 in ~/.warden/.env file\033[0m"
+
     export MUTAGEN_SYNC_FILE="${WARDEN_DIR}/environments/${WARDEN_ENV_TYPE}/${WARDEN_ENV_TYPE}.mutagen.yml"
 
     if [[ -f "${WARDEN_HOME_DIR}/environments/${WARDEN_ENV_TYPE}/${WARDEN_ENV_TYPE}.mutagen.yml" ]]; then
--- a/commands/install.cmd
+++ b/commands/install.cmd
@@ -93,5 +93,7 @@ if [[ ! -f "${WARDEN_HOME_DIR}/.env" ]]; then
 		WARDEN_PORTAINER_ENABLE=0
 		# SEt to "0" to disable DNSMasq
 		WARDEN_DNSMASQ_ENABLE=1
+        # Set to "1" to enable Mutagen
+        WARDEN_MUTAGEN_ENABLE=1
 	EOT
 fi
--- a/utils/env.sh
+++ b/utils/env.sh
@@ -53,6 +53,10 @@ function loadEnvConfig () {
         ;;
     esac
 
+    if [[ -f "${WARDEN_HOME_DIR}/.env" ]]; then
+      eval "$(sed 's/\r$//g' < "${WARDEN_HOME_DIR}/.env" | grep "^WARDEN_")"
+    fi
+
     assertValidEnvType
 }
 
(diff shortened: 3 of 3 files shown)


lewisvoncken (approved):

commands/env.cmd
```diff
@@ -19,6 +19,11 @@ trap '' ERR
 ## define source repository
 if [[ -f "${WARDEN_HOME_DIR}/.env" ]]; then
   eval "$(sed 's/\r$//g' < "${WARDEN_HOME_DIR}/.env" | grep "^WARDEN_")"
+
+  ## Temporary fix to force mutagen activation
+  if [[ ${WARDEN_MUTAGEN_ENABLE} == '' ]]; then
```
bap14, line 24: @joseluisi4 We should follow the previous examples of providing a default for a variable that may be empty.

`WARDEN_MUTAGEN_ENABLE=${WARDEN_MUTAGEN_ENABLE:-1}`

utils/env.sh
```diff
@@ -122,18 +126,28 @@ function appendEnvPartialIfExists () {
...
         "${WARDEN_ENV_PATH}/.warden/environments/includes/${PARTIAL_NAME}.${WARDEN_ENV_SUBT}.yml" \
-        "${WARDEN_ENV_PATH}/.warden/environments/${WARDEN_ENV_TYPE}/${PARTIAL_NAME}.base.yml" \
-        "${WARDEN_ENV_PATH}/.warden/environments/${WARDEN_ENV_TYPE}/${PARTIAL_NAME}.${WARDEN_ENV_SUBT}.yml"
+        "${WARDEN_ENV_PATH}/.warden/environments/${WARDEN_ENV_TYPE}/${PARTIAL_NAME}.base.yml"
     do
         if [[ -f "${PARTIAL_PATH}" ]]; then
             DOCKER_COMPOSE_ARGS+=("-f" "${PARTIAL_PATH}")
         fi
     done
+
```
bap14, line 140: @joseluisi4 This doesn't seem like it would work to me. The value of `PARTIAL_NAME` will be "mutagen", but there's no check against the partial name.

I think a better solution would be to check the `PARTIAL_NAME` prior to the for loop, and if it's "mutagen" and mutagen is disabled, then just return from the function; otherwise, process the for-loop as originally developed.

bap14 (changes requested):

utils/env.sh
joseluisi4: `PARTIAL_NAME` will never be "mutagen", it is the name of the **environment**

The problem I am trying to solve is that the `xxx.darwin.yml` files are not used because these volumes are designed for Mutange

Another solution would be to rename those files to `xxx.mutagen.yml`, this way you could filter by name

bap14: Partial name is _**not**_ the name of the environment. The `appendEnvPartialIfExists` function takes the name of the service partial to include (e.g. `nginx`, or `php`), not the environment. See https://github.com/wardenenv/warden/blob/39d472934585b7a656033f52c5ae4150470912bd/commands/env.cmd#L80

My reservation about removing the OS-specific partials is that if you have an os-specific partial for a service in your environment but aren't using Mutagen, it will never be included.

To me this would be an opportunity to properly refactor the mutagen inclusion to use the `appendEnvPartialIfExists` function to add Mutagen syncs, and only including them if the OS is darwin and Mutagen is enabled. This would then also mean that the checks for starting / pausing Mutagen would need to be adjusted to use the `WARDEN_MUTAGEN_ENABLE` variable.

commands/env.cmd
```diff
@@ -174,7 +177,10 @@ TRAEFIK_ADDRESS="$(docker container inspect traefik \
 )"
 export TRAEFIK_ADDRESS;
 
-if [[ $OSTYPE =~ ^darwin ]]; then
+if [[ $OSTYPE =~ ^darwin ]] && [[ ${WARDEN_MUTAGEN_ENABLE} -eq 1 ]]; then
+    echo -e "\033[31mMutagen is being deprecated!!\033[0m"
```
ihor-sviziev, line 181: Did anyone compare performance on different Docker installations, not only for OrbStack?
Unless we confirm that it works better w/o mutagen all the time, we can't make it deprecated. It will add confusion

joseluisi4: Hello @ihor-sviziev, in the latest versions of Docker Desktop, performance is higher, RAM consumption is reduced and we will not have file synchronization problems.
It is important to use VirtioFS

bap14: @joseluisi4 I think what @ihor-sviziev is getting at is: do we have numbers to back up the fact that Docker bind mounts are now more performant than Mutagen? Docker can spout off that it fixed memory or resource usage, but it doesn't always turn into real-world gains in every scenario.

So the question remains is do we have evidence to back up the claim that docker has higher performance, less RAM consumption and no file sync problems? Do we have it just for a single system type, or is it across all system types (Windows, Intel Mac, ARM Mac, x86_64 Linux, ARM Linux)? Without evidence to that, I think he's suggesting we maybe not mark it as deprecated yet, and wait until we get more concrete evidence that it is in fact more performant before calling it deprecated.

ihor-sviziev: @bap14, yeah, you're right. My suggestion was not to mark it as deprecated but rather add the ability to disable it in a backward-compatible manner.
We might add a notice like "We noticed that you're on MacOS and using Mutagen sync. We recommend configuring Docker Desktop to use VirtioFS and disable mutagen sync by adding WARDEN_MUTAGEN_DISABLED=1 to your project .env file.".
Also, I have a question: how will it work for those now using mutagen but then switching to non-mutagen installation? We should probably recommend terminating some volume to reduce disk usage. This also should be documented.

In the next six months, we can collect feedback if people see any issues by disabling mutagen on macOS. There might be cases when mutagen-enabled flow works significantly faster. If no such issues - we can finally mark it as deprecated

dudzio12:
I have other thoughts on this topics. I have my .env files saved to git repositories and all team is using it. I have no control over environments of other people, so some might be using OrbStack, VirtioFS and some not. Since there is no global warden config, other then project env, I would suggest to implement behavior to check if mutagen is installed and run sync if it is.

joseluisi4:
> I have other thoughts on this topics. I have my .env files saved to git repositories and all team is using it. I have no control over environments of other people, so some might be using OrbStack, VirtioFS and some not. Since there is no global warden config, other then project env, I would suggest to implement behavior to check if mutagen is installed and run sync if it is.

Hi @dudzio12 the `.env` we are referring to is not the projects `.env`, it is a Warden configuration `.env`, it is in `~/.warden/.env`

commands/env.cmd
Den4ik: I'm agree with @ihor-sviziev
I have checked performance on mac M1 Pro and my subjective assessment - there is no performance difference.
Make sense to wait another developers responses and to have this option.

dudzio12:
> > I have other thoughts on this topics. I have my .env files saved to git repositories and all team is using it. I have no control over environments of other people, so some might be using OrbStack, VirtioFS and some not. Since there is no global warden config, other then project env, I would suggest to implement behavior to check if mutagen is installed and run sync if it is.
> 
> Hi @dudzio12 the `.env` we are referring to is not the projects `.env`, it is a Warden configuration `.env`, it is in `~/.warden/.env`

Thanks for the clarification :) Looking forward for the warden update 👍

Den4ik:
I believe that mutagen volumes configuration should be moved to separate file, because we can have configs directly related to platform.
For example, we may use `platform: ${DOCKER_PLATFORM:-linux/amd64}` platform related config in MacOS. It's allow to run non ARM images on the Apple Chip. https://github.com/magenius-team/warden/commit/6ecb5eb038fddf475fecd09f0709fa913e7ac81d

dudzio12:
@joseluisi4 Do you plan to fix issues mentioned by @navarr and @bap14? I've tested warden without mutagen for few months then upgraded warden and the speed is only minor feature for me. Lack of mutagen, which frequently causes issues with sync ex. strange symlinks, missing files etc. is much more important for me personally. If you do not have time to finish your PR, let us now, so I'll fix it and hopefully it could be released in next warden version :)

ihor-sviziev:
@joseluisi4, just wonder, what's left on this PR?

diazwatson:
> @joseluisi4, just wonder, what's left on this PR?

Maybe @joseluisi4 gave up after almost one year trying to get this PR merged :P

bap14:
I'm not sure what happened to them, but if anyone wants to pick this up and fix the conflicts and incorporate the feedback Navarr and I have given, we can move forward with it.

ihor-sviziev:
I opened a new PR https://github.com/wardenenv/warden/pull/842 that fixes listed above issues

commands/env.cmd
```diff
@@ -19,6 +19,9 @@ trap '' ERR
 ## define source repository
 if [[ -f "${WARDEN_HOME_DIR}/.env" ]]; then
   eval "$(sed 's/\r$//g' < "${WARDEN_HOME_DIR}/.env" | grep "^WARDEN_")"
+
+  ## configure mutagen enable by default
+  WARDEN_MUTAGEN_ENABLE=${WARDEN_MUTAGEN_ENABLE:-1}
```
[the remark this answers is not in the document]
ihor-sviziev: Fixed in https://github.com/wardenenv/warden/pull/842/commits/5dc2fa5f4ede1525020f0f583f2703eff750c5d6

commands/install.cmd
```diff
@@ -93,5 +93,7 @@ if [[ ! -f "${WARDEN_HOME_DIR}/.env" ]]; then
 		WARDEN_PORTAINER_ENABLE=0
 		# SEt to "0" to disable DNSMasq
 		WARDEN_DNSMASQ_ENABLE=1
+        # Set to "1" to enable Mutagen
+        WARDEN_MUTAGEN_ENABLE=1
```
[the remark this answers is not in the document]
ihor-sviziev: It was already fixed

PR #750, opened 2024-02-19
Implement Ngrok command to use with Nginx

anbis:


commands/ngrok.cmd
bap14, line 1: Why use this approach instead of using another yaml config file? It could then be configured via the `.env` file.

Here's an example yaml snippet that achieves the same result:
```yaml
services:
  ngrok:
    image: ngrok/ngrok:latest
    entrypoint:
      - ./entrypoint.sh
      - http
      - nginx:80
    environment:
      - NGROK_AUTHTOKEN=${WARDEN_NGROK_TOKEN}
    links:
      - nginx
    networks:
      - default
```

anbis: Just because ngrok - this is tool, that not really needed so often, I decided to do it as command, just because to don't restart project when u decided to check something in real Internet, to don't waste time to restart project. BTW in case if I need to do it as separate yml to implement support of ngrok to warden - I will do it, just say me preferable way.

navarr: I agree with this approach as a command.  Opening up your local network to the internet is probably not best to do automatically on project startup.

commands/ngrok.cmd
```diff
@@ -0,0 +1,22 @@
...
+if [[ ! ${NGINX_CONTAINER} ]]; then
+    fatal "No container found for nginx service."
+fi
+
+NETWORK_NAME=$(docker container inspect ${NGINX_CONTAINER} --format '{{range $key, $value := .NetworkSettings.Networks}}{{printf "%s\n" $key}}{{end}}')
+if [[ ! ${NETWORK_NAME} ]]; then
+    fatal "No networks found for nginx service."
+fi
+
+docker run --rm -it --link ${NGINX_CONTAINER} --net ${NETWORK_NAME} -e NGROK_AUTHTOKEN=${WARDEN_NGROK_TOKEN} ngrok/ngrok:latest http ${NGINX_CONTAINER}:80
```
navarr, line 22: I think this would be better if the `NGROK_AUTHTOKEN` was checked as a parameter first, and then fall back to `WARDEN_NGROK_TOKEN`.  This would better support one-off instantiations.

For bonus points, we could offer to save the token to the `~/.warden/.env` file - but I'm not going to ask you to do that amount of work if you don't want to.

I would then suggest that `WARDEN_NGROK_TOKEN` be set in the `~/.warden/.env` file instead of the project file - as unlike Blackfire, Ngrok authentication is more likely to be per-user rather than per-project.

environments/magento2/init.env
```diff
@@ -7,6 +7,8 @@ WARDEN_VARNISH=1
 WARDEN_RABBITMQ=1
 WARDEN_REDIS=1
 
+WARDEN_NGROK_TOKEN=
+
```
navarr, line 11: Based on my other note, I think this file shouldn't change.

navarr (changes requested):

commands/ngrok.cmd
anbis: will do

navarr: @anbis Appreciate your efforts here <3

navarr (approved):
Looks great!  Pending QA

anbis:
@navarr do I need to prepare something for QA?

navarr:
Closing this out in favor of #769

PR #751, opened 2024-02-20
Implement Sablier Traefik Plugin for Workload Management

SamJUK:
## Overview
Implement the [Sablier Traefik Plugin](https://plugins.traefik.io/plugins/633b4658a4caa9ddeffda119/sablier), allowing Warden to dynamic start and stop project containers on request rather than having to manage them with env up/down saving system resources for us forgetful people. Especially useful if your working on multiple large projects simultaneously, or for certain edge use cases (we run Warden also as a Ephemeral/Feature hosts deployment with a lot of projects for example).

## Implementation Notes
- Disabled by default, can be enabled via setting `WARDEN_STABLIER_ENABLE =1` in the service `.env`
- With stop all containers except the ingress (varnish/nginx depending on project configuration)
- Session lifetime can be set per project via `TRAEFIK_SABLIER_DURATION=2h` (How long before Sablier sleeps the containers)
- Custom warden theme (Good PR opportunity for someone to make it look decent)

## Configuration
Service Config:
`WARDEN_STABLIER_ENABLE=0` - Enables Sablier globally for the installation (Unable to think of a good use case of enablement on project basis)

Project :
`TRAEFIK_SABLIER_DURATION=30m` - How long Sablier waits to sleep containers
`TRAEFIK_SABLIER_THEME=ghost` - Sablier theme to use [List of Default Themes](https://acouvreur.github.io/sablier/#/themes)

Advanced:
`${WARDEN_HOME_DIR}/etc/sablier/theme` - Directory if you want to declare any custom Sablier Themes
`${WARDEN_HOME_DIR}/etc/sablier/sablier.yml` - Sablier Service Configuration Overrides

### DIFF
--- a/config/sablier/sablier.yml
+++ b/config/sablier/sablier.yml
@@ -1,0 +1,20 @@
+provider:
+  name: docker
+server:
+  port: 10000
+  base-path: /
+storage:
+  file:
+sessions:
+  default-duration: 5m
+  expiration-interval: 20s
+logging:
+  level: info
+strategy:
+  dynamic:
+    custom-themes-path:
+    show-details-by-default: true
+    default-theme: warden
+    default-refresh-frequency: 5s
+  blocking:
+    default-timeout: 1m
--- a/config/traefik/traefik.yml
+++ b/config/traefik/traefik.yml
@@ -23,3 +23,8 @@ log:
 global:
   checkNewVersion: false
   sendAnonymousUsage: false
+experimental:
+  plugins:
+    sablier:
+      moduleName: github.com/acouvreur/sablier
+      version: v1.5.0
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -9,7 +9,7 @@ services:
     volumes:
       - ${WARDEN_HOME_DIR}/etc/traefik/traefik.yml:/etc/traefik/traefik.yml
       - ${WARDEN_HOME_DIR}/etc/traefik/dynamic.yml:/etc/traefik/dynamic.yml
-      - ${WARDEN_HOME_DIR}/ssl/certs:/etc/ssl/certs
+      - ${WARDEN_HOME_DIR}/ssl/certs:/etc/ssl/certs/warden
       - /var/run/docker.sock:/var/run/docker.sock
     labels:
       - traefik.enable=true
--- a/commands/env.cmd
+++ b/commands/env.cmd
@@ -125,6 +125,13 @@ fi
 [[ ${WARDEN_MAGEPACK} -eq 1 ]] \
     && appendEnvPartialIfExists "${WARDEN_ENV_TYPE}.magepack"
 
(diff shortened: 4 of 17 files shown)


docker/docker-compose.yml
```diff
@@ -9,7 +9,7 @@ services:
     volumes:
       - ${WARDEN_HOME_DIR}/etc/traefik/traefik.yml:/etc/traefik/traefik.yml
       - ${WARDEN_HOME_DIR}/etc/traefik/dynamic.yml:/etc/traefik/dynamic.yml
-      - ${WARDEN_HOME_DIR}/ssl/certs:/etc/ssl/certs
+      - ${WARDEN_HOME_DIR}/ssl/certs:/etc/ssl/certs/warden
```
navarr, line 12: The SSL certificate location change is outside the scope of your PR

navarr (changes requested):

docker/docker-compose.yml
SamJUK: The original mount ends up clobbering all the default system certificates. Preventing Traefik from downloading  plugins due to SSL verification errors.
Could be split to a seperate PR, as it should affecting install any Traefik plugin

config/sablier/sablier.yml
```diff
@@ -0,0 +1,20 @@
+provider:
+  name: docker
+server:
+  port: 10000
+  base-path: /
+storage:
+  file:
+sessions:
+  default-duration: 5m
```
Den4ik, line 9: I believe that this config should be configurable, because you could work on the code and don't use env much more time. You will lose console history if it will be closed by stoping environment

docker/docker-compose.yml
bap14: @SamJUK To @navarr 's point, then there should be a separate PR that fixes the clobbering issue, and this one requires that one to be fixed first. The PR should do a single thing (add Sablier) not multiple (add Sablier, change SSL dirs, etc.)

SamJUK: Fair points, separated this change to PR #812

config/sablier/sablier.yml
SamJUK: Agreed, implemented the same way as `traefik.yml` for consistency. It can be configured via `${WARDEN_HOME_DIR}/etc/sablier/sablier.yml`, Although it will be overwritten on `svc up` calls. 

Opened PR #813 which looks to migrate the Traefik config to environment variables. Where users can override / expand the configuration via the `~/.warden/docker-compose.yml` file. Sablier can be refactored to the same approach if approved.

config/traefik/traefik.yml
```diff
@@ -23,3 +23,8 @@ log:
 global:
   checkNewVersion: false
   sendAnonymousUsage: false
+experimental:
```
SamJUK, line 26: If PR #813 is approved. This can be refactored into `docker/docker-compose.sablier.yml` as environment variables. Preventing the need to install the plugin even if Sablier is disabled.

PR #754, opened 2024-02-26
Feature/add postgresql support

bap14:
After adding the PostgreSQL images, this adds support to warden environments (or should). This emits a warning that using partials like `db.base.yml` or `db.darwin.yml` are deprecated and should be replaced with `db.mysql.base.yml` or `db.mysql.darwin.yml`.

I tested the DB connection and DB dump commands for PostgreSQL and they worked.


bap14:
The default is still mysql / mariadb. It adds a new `WARDEN_DB_SYSTEM` variable that is either mysql / mariadb or postgres. It will include the proper partial(s).

The current drawback is that only one database is allowed at a time, so you can't run both mysql and postgres. That's a feature that could be done in the future, but at the moment the key was to get it functioning.

navarr:
Thanks for your hard work here, @bap14! Looks like I have a lot to review tomorrow.

I will likely change or request changed the name to match a wip PR I have to align env variables with Den.

But I'm excited to take a deeper look at this tomorrow

alikhani97:
any update on this ?!

PR #769, opened 2024-05-28
feat(ngrok): Adding ngrok feature

bruno-blackbird:
Hello,

There is a working implementation of ngrok with caddy as a reverse proxy to automatically map urls from : 

https://domain1.test to https://497b85f07962.ngrok.app
https://domain2.test to https://5fc752cab8e5.ngrok.app

It avoids having to change any url configuration in your local environement, the only limitation I found from now is that it does not work with obfuscated urls.

I also created a pull request on wardenenv/images and wardenenv/docs


navarr:
Great first pass!  I'm working on some modifications in branch `feature/warden-769-ngrok`. The most important of those is that I would like to enable NGROK_AUTHTOKEN being read from the home directory ~/.warden/.env.  These tokens are likely unique per developer as opposed to per-project, so unlike most warden configuration one would not want them in vcs.

bruno-blackbird:
Great :) 
I understand the point, but I already get the case where it was versioned for a project using. But it is clearly not the most common use case.
We had a premium account with a dedicated domain name. It was useful as we were working with a Global-e deported checkout integration that needed a dedicated url for the environment configuration.
We were launching ngrok depending on who needed to work with the integration. It can also be the case for some payment methods, where you cannot easily modifiy the external url.

environments/includes/ngrok.base.yml
```diff
@@ -0,0 +1,26 @@
+services:
+  ngrok:
+    hostname: "${WARDEN_ENV_NAME}-ngrok"
+    image: ngrok/ngrok:${NGROK_VERSION:-latest}
+    environment:
+      NGROK_AUTHTOKEN: ${NGROK_AUTHTOKEN}
+    volumes:
+      - ${WARDEN_ENV_PATH:-.}/.warden/ngrok.yml:/etc/ngrok.yml:ro,cached
+    command:
```
bap14, line 9: @bruno-blackbird Reviewing the entrypoint script in the ngrok image, I think all you'd need to do here is just specify the `NGROK_CONFIG` environment variable to be `/etc/ngrok.yml`. The entrypoint script will start listening on all tunnels in the config file by default.

bap14:
I tried to manually do this locally, and I'm wondering about the caddy config. It seems it's only adjusting "get" requests. What about POST? Will it work to add an item to the cart? Or log in? Or sign up? I'm not super familiar with Caddy and its syntax, so I'm just asking.

environments/includes/ngrok.base.yml
bruno-blackbird: @bap14 you mean instead of passing the command, yes, maybe it will be cleaner if I  add it this way

bruno-blackbird:
@bap14 Yes it does replace only in @get, which is enough for me, I never get any need to replace urls in POST  requests as adding item to the cart, log in and sign up are working fine.

The only issue I get is with website with obfuscation on urls, as they are base 64 encoded, maybe I should add base_64 version of urls in caddy config too.

environments/includes/ngrok.base.yml
bap14: @bruno-blackbird Yes, that's correct.

PR #773, opened 2024-06-06
Add environment version support.

hardyjohnson:
This PR adds versioning of the prebuilt environment configurations as per discussion here: https://github.com/orgs/wardenenv/discussions/772#discussion-6782904

## Environment versioning

In order to more accurately and effectively support the modeling of an application's services and their configurations,
an "environment" within Warden, we should add a versioning methodology to each supported environment.

Adding an environment versioning constraint will allow us to support accurate representation of the necessary services and configuration necessary to run a particular application and support it over time.
Current issues with the pre-built environments today:

- It is difficult to tell what version of an application works with each pre-built environment included within wardenenv/warden.
- No support for environment upgrades based on an application version e.g. upgrade current installation from Magento2 2.4.5 to 2.4.7 service stack.

## Proposed solution:

    Add a convention for environment service versioning
        ./environments/WARDEN_ENV_TYPE/{WARDEN_ENV_VERSION:-DEFAULT} - this path would contain the .init.env, base.yml and docker compose environment partials for a particular application version.
    Add an application version variable to the .env when using env-init to represent the supported application's modelled service and configuration version.
        WARDEN_ENV_TYPE=magento2
        WARDEN_ENV_VERSION=2.4.7
    Update the env-init and env commands to work with these changes, allow for application version to be passed in env-init.
    Update documentation explaining application version usage.

## Usage

Environment initialization: `warden env-init newapp magento2 2.4.7`

This PR includes environments Magento2 (2.45, 2.4.6, 2.4.7) as per https://experienceleague.adobe.com/en/docs/commerce-operations/release/versions on premise commerce versions table.

cc/ @bap14 @navarr


commands/env-init.help
```diff
@@ -3,7 +3,7 @@
 
 WARDEN_USAGE=$(cat <<EOF
 \033[33mUsage:\033[0m
-  env-init <project_name> <environment_type>
+  env-init <project_name> <environment_type> <environment_version>
```
bap14, line 6: Since `<environment_version>` is optional, it should be surrounded by square brackets: `[<environment_version>]`; otherwise, it appears to be a required argument.

environments/magento2/default/magento2.base.yml
bap14, line 1: There are a lot of files that may be exact copies. I feel it would be better if we adjusted the environment files to use a fallback mechanism. If a file doesn't exist in the specific version, it loads from `default`. This would simplify the environment releases to potentially just `init.env` files.

bap14:
@hardyjohnson You'll want to resolve the conflicts that exist as well.

PR #779, opened 2024-06-25
Add optional (opt-in) support for Mailpit

bap14:
Adds ability to easily switch to Mailpit, a maintained successor to Mailhog.


bap14:
@navarr This goes for this discussion: https://github.com/orgs/wardenenv/discussions/780

docker/docker-compose.mailpit.yml
bap14, line 1: The tags in this file don't use the "array of strings" format because the redirect wasn't working unless I used the object style notation.

bap14:
@navarr This is ready for your review.

navarr (approved):

docker/docker-compose.mailpit.yml
```diff
@@ -0,0 +1,27 @@
+services:
+  mailpit:
+    container_name: mailhog
```
ihor-sviziev, line 3: why the container name is mailhog?

bap14: For backwards compatibility so you can still reference "mailhog" but it's really just mailpit. This can easily be removed and whatever breaks will need to be upgraded.

bap14: The simple solution would be: `warden svc down --remove-orphans && warden svc up`.

ihor-sviziev: Yeah, I did like that and it worked. But I got confused that container is named mailhog, while we don't have such service and domain. So I would prefer removing this line

PR #787, opened 2024-08-01
Add hyva back to sponsors

wigman:
Hyvä has sponsored twice the amount Sansec has sponsored (and continues to sponsor till date), so it seems fair to still be mentioned as Gold Sponsor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,8 @@ This work is licensed under the MIT license. See [LICENSE](https://github.com/da
 This project was started in 2019 by [David Alger](https://davidalger.com/).
 
 # Gold Sponsors
+[![Hyvä](https://user-images.githubusercontent.com/145128/226427529-53483968-c9ab-484a-9ae3-c6abb58f81c9.png)](https://www.hyva.io/)
+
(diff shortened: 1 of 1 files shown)


wigman:
For reference: https://opencollective.com/warden

@navarr

bap14:
@wigman  According to Open Collective, we only have "Backers" and "Supporters", and only "Supporters" have the benefit of being listed on the Warden site. They're also listing Hyvä as a "Backer" level, not "Supporter". That's the only reason I am marking this as rejected, but @navarr can chime in and override it if he feels.

navarr:
@bap14 I've spoken directly to Willem about this. I'd like to leave this open while I think of ways of honoring previous sponsors. Hyvå was incredibly lovely and helpful as was Swiftotter and they deserve some amount of continuing recognition (that doesn't detract the current sponsors)

bap14:
@navarr I figured there was a history there I wasn't aware of.

ihor-sviziev:
Just an idea - what if add to sponsors not only current sponsors, but also those who donated let's say >$500 over last year?

PR #800, opened 2024-09-16
feat(elk): add elk-based observability tools stack (kibana + vector)

theuargb:
This PR aims to provide tools to enable log overviews against the environment and allow developers to experiment with observability tech for free.

It is DRAFT since I need to push the corresponding PR into the images repo and polish some configs. Still, I want the community to be able to start a discussion as early as possible.


theuargb:
the problem is that kibana and its opensearch-ed variant requires versions to be nearly equal with the engine itself (7.17 kibana wont start with the 7.8 elastic, etc.), so (as far as I see) it would be impossible to have one service for all the envs(

Issue #540, opened 2022-08-29
[SUPPORT REQUEST] Traefik ACME configuration


manuelcanepa:
Im trying to setup the acme configuration to auto-sign with let's encrypt. After editing my docker-compose.yml (#531) and starting the environment, traefik logs shows an error:
```
level=error msg="the router homeassistant@docker uses a non-existent resolver: myresolver"
```

I follow all the steps on traefik documentation and still getting this error. 

`/opt/warden/docker/docker-compose.yml`:

```yml
[...]
    container_name: traefik
    image: traefik:${TRAEFIK_VERSION:-2.2}
    command:
      - "--log.level=info"
      - "--api.insecure=true"
      - "--entrypoints.web.address=:80"
      - "--entrypoints.websecure.address=:443"
      - "--certificatesresolvers.myresolver.acme.caserver=https://acme-v02.api.letsencrypt.org/directory"
      - "--certificatesresolvers.myresolver.acme.tlschallenge=true"
      - "--certificatesresolvers.myresolver.acme.email=....."
      - "--certificatesresolvers.myresolver.acme.storage=acme.json"
      - "--certificatesresolvers.myresolver.acme.httpchallenge=true"
      - "--certificatesresolvers.myresolver.acme.httpchallenge.entrypoint=web"
[...]

```

Has anyone been able to resolve this issue?

navarr:
Can you show your homeassistant@docker config?

manuelcanepa:
> Can you show your homeassistant@docker config?

Yes, of course:

```
  homeassistant:
    image: ghcr.io/home-assistant/home-assistant:stable
    privileged: true
    restart: unless-stopped
    environment:
      - TZ=America/Argentina/Buenos_Aires
    volumes:
      - ./config:/config
      - /etc/localtime:/etc/localtime:ro
      - /var/run/dbus/:/run/dbus/:z
    expose:
      - 8123

    labels:
      # DUCKDNS
      - traefik.enable=true
      - traefik.http.routers.${WARDEN_ENV_NAME}.tls=true
      - traefik.http.routers.${WARDEN_ENV_NAME}.tls.certresolver=myresolver
      - traefik.http.routers.${WARDEN_ENV_NAME}.priority=4
      - traefik.http.routers.${WARDEN_ENV_NAME}.rule=Host(`privatedomain.duckdns.org`)
      # LOCAL DOMAIN
      - traefik.http.routers.${WARDEN_ENV_NAME}-local.tls=true
[...]

PR #813, opened 2024-10-20
Migrate Traefik Configuration To Environment Variables

SamJUK:
Currently overriding / extending the `traefik.yml` configuration is difficult. This PR proposes migrating the Traefik configuration from the `.yml` file to environment variables. This should improve both the extendability and maintainability of attempting to extend the Traefik config. Such as, but not limited to configuring plugins or certificate resolvers.

I've found there are 2 main pain points to the current approach.

1. You have to overwrite and maintain the whole configuration file, even for a single value change. 
2. You have to edit the source file, since the `svc up` command replaces any modifications under `$HOME/.warden/etc/traefik/traefik.yml` [See Here](https://github.com/wardenenv/warden/blob/main/commands/svc.cmd#L65)

Note: May be considered a breaking change. Since any custom changes to the `traefik.yml` by users would have to be reimplemented via the ENV approach. Due to [Traefik static config approaches being mutually exclusive](https://doc.traefik.io/traefik/getting-started/configuration-overview/#the-static-configuration). 

---

## Example Use Case

Say you want to enable debug logging for Traefik. With the environment variable approach you can create a docker compose file in the warden home directory with the following contents. 
```yml
# ~/.warden/docker-compose.yml
services:
  traefik:
      TRAEFIK_LOG_LEVEL: DEBUG
```

Or if you want to define a certificate resolver such as LetsEncrypt to sign your certs. It can again be defined in the users home directory with the following docker compose file. [See this issue of someone attempting a similar thing and having to override the source file](https://github.com/wardenenv/warden/issues/540#issuecomment-1281177584)
```yml
# ~/.warden/docker-compose.yml
services:
  traefik:
    volumes:
      - acmedata:/acme
    environment:
      CF_DNS_API_TOKEN: xxx
      CF_ZONE_API_TOKEN: xxx
      TRAEFIK_CERTIFICATESRESOLVERS_LETSENCRYPT: true
      TRAEFIK_CERTIFICATESRESOLVERS_LETSENCRYPT_ACME_EMAIL: admin@example.com
      TRAEFIK_CERTIFICATESRESOLVERS_LETSENCRYPT_ACME_STORAGE: /acme/acme.json
      TRAEFIK_CERTIFICATESRESOLVERS_LETSENCRYPT_ACME_CASERVER: https://acme-staging-v02.api.letsencrypt.org/directory
      TRAEFIK_CERTIFICATESRESOLVERS_LETSENCRYPT_ACME_DNSCHALLENGE: true
      TRAEFIK_CERTIFICATESRESOLVERS_LETSENCRYPT_ACME_DNSCHALLENGE_PROVIDER: cloudflare
volumes:
  acmedata:
```

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -6,7 +6,6 @@ services:
       - "${TRAEFIK_LISTEN:-127.0.0.1}:80:80"     # The HTTP port
       - "${TRAEFIK_LISTEN:-127.0.0.1}:443:443"   # The HTTPS port
     volumes:
-      - ${WARDEN_HOME_DIR}/etc/traefik/traefik.yml:/etc/traefik/traefik.yml
       - ${WARDEN_HOME_DIR}/etc/traefik/dynamic.yml:/etc/traefik/dynamic.yml
       - ${WARDEN_HOME_DIR}/ssl/certs:/etc/ssl/certs
       - /var/run/docker.sock:/var/run/docker.sock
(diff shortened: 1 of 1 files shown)


bap14:
Thank you for the work on this @SamJUK !! I see no issues with this PR.

If we didn't want to wait until Warden 1.0 (or 0.15) to push this change we could add a Warden flag that is `WARDEN_USE_TRAEFIK_ENV_CONFIG` that if set to `1` then the config file is this new style, but if not set is the old way. Then also send a message to `stderr` which states that the Traefik static file config has been deprecated and will be removed in a future version and people should start switching to `WARDEN_USE_TRAEFIK_ENV_CONFIG` now.

Thoughts @navarr ?

hardyjohnson:
I believe there is a happy path to an alternative with similar functionality which would not cause breaking change.

This could be achieved by creating a method which appends a partial from the home `~/.warden/config/traefik/traefik.yml` during svc up,  similar to how this is done for environments partials. This would follow the established pattern for extension and overriding without the need to break any existing setups.

If this suggestion is ignored and this PR moves forward so be it, but it should be a complete PR including the following:

1. Parallel PR in the docs repo linked to this PR explaining the change for global service override for Traefik.
2. Remove the existing `config/traefik/traefik.yml` file.
3. Remove the copying of the traefik.yml file during `svc up`.

PR #820, opened 2024-11-17
feat: PHP SPX

SamJUK:
Supporting work to make use of the new PHP SPX images introduced in https://github.com/wardenenv/images/pull/17

PR for Supporting documentation: https://github.com/wardenenv/docs/pull/25

**Key Points:**
- New boolean feature flag `WARDEN_PHP_SPX` to toggle feature state
- Feature flag added to each .env init template, with a default of disabled
- Update `warden shell` to automatically pick the correct container depending on the feature flag


commands/shell.cmd
```diff
@@ -7,7 +7,9 @@ loadEnvConfig "${WARDEN_ENV_PATH}" || exit $?
 ## set defaults for this command which can be overridden either using exports in the user
 ## profile or setting them in the .env configuration on a per-project basis
 WARDEN_ENV_SHELL_COMMAND=${WARDEN_ENV_SHELL_COMMAND:-bash}
-WARDEN_ENV_SHELL_CONTAINER=${WARDEN_ENV_SHELL_CONTAINER:-php-fpm}
+
+WARDEN_ENV_DEFAULT_SHELL_CONTAINER=$([ $WARDEN_PHP_SPX == 1 ] && echo "php-spx" || echo "php-fpm")
```
bap14, line 11: @SamJUK I don't fully agree with this. The `warden shell` goes to the standard php-fpm container and `warden debug` goes to php-debug container. To me this would seem like we could either add a flag `--spx` or create a new `warden spx` command to drop to a shell in it.

SamJUK: Makes sense, I'll switch it over to a separate command `warden spx` (since AFAIK the flag approach is not widely used across the project)

ihor-sviziev (approved):

ihor-sviziev:
Waiting for these changes to be merged and delivered as a release that we can use on daily basis :)

monteshot:
I updated Warden (alternative installation)
Added to project file(.env):
```
WARDEN_PHP_SPX=1
```

Then ran 
```warden svc pull; warden svc up; warden env pull; warden env up;```  
 
I successfully get SPX UI 
image.

Then I disabled SPX

```WARDEN_PHP_SPX=0```

And run
```
warden env down; warden env up;
``` 

The Project is not working anymore
image
image

SamJUK:
Can you make sure you do not have any SPX related cookies remaining? Or try in Incognito / with another browser?

ihor-sviziev:
I guess it has to be added to documentation https://github.com/wardenenv/docs/pull/25#issuecomment-2490963821

monteshot:
I tried to Disable SPX from the UI... I think there is an issue because SPX cookies are still present.

> "Another option - remove SPX_* cookies from your browser"

This solution works as a workaround. I think an additional mechanism(on Traefik or somewhere else) should be created only for disabling the feature.

bap14:
@SamJUK Unfortunately there are going to end up being manual steps no matter what because disabling the service won't update the cookies in your browser. Just like if you stopped the xdebug container and then sent a request with the `XDEBUG_SESSION` cookie set it would also result in an error. To my knowledge there isn't a PHP-SPX chrome extension, but that could be a potential solution for this, but that's also technically a manual process as well.

Issue #823, opened 2024-11-26
PhpMyAdmin - error is shown after updating warden to dev version


ihor-sviziev:
Follow up to https://github.com/wardenenv/warden/pull/801:

> Today, I installed dev version of Warden, ran `warden svc up`, and got the following notice:
<img width="1637" alt="image" src="https://github.com/user-attachments/assets/36cd6e5f-8547-46bf-8757-00537033688e">

Also, running `warden svc config` shows an error:
<img width="1246" alt="image" src="https://github.com/user-attachments/assets/058db27b-9da0-4f38-bcf6-eaeeabda6ecc">

@monteshot could you check that?

_Originally posted by @ihor-sviziev in https://github.com/wardenenv/warden/issues/801#issuecomment-2498642767_

monteshot:
It seems like there is no Docker image for PMA. I don't have a Mac. I will try to reproduce it on WSL.

monteshot:
Clear WSL Ubuntu:
![image](https://github.com/user-attachments/assets/79ca9b2b-dab5-437e-b4c2-01ee55b79503)

@ihor-sviziev, What is the best way to resolve this issue?

monteshot:
Mb this PR will fix the root issue with incorrect directory/file creation during the docker-compose spins up the project
https://github.com/wardenenv/warden/pull/824

PR #824, opened 2024-11-26
fix(svc.cmd): ensure phpMyAdmin config file is consistently managed

monteshot:
Add checks to remove outdated phpMyAdmin config file if it exists and create a new one if absent. Clean up redundant directory creation steps.

https://github.com/wardenenv/warden/issues/823

### DIFF
--- a/commands/svc.cmd
+++ b/commands/svc.cmd
@@ -45,6 +45,13 @@ fi
 
 WARDEN_PHPMYADMIN_ENABLE="${WARDEN_PHPMYADMIN_ENABLE:-1}"
 if [[ "${WARDEN_PHPMYADMIN_ENABLE}" == 1 ]]; then
+    if [[ -d "${WARDEN_HOME_DIR}/etc/phpmyadmin/config.user.inc.php" ]]; then
+        rm -rf ${WARDEN_HOME_DIR}/etc/phpmyadmin/config.user.inc.php
+    fi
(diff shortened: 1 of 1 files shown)


commands/svc.cmd
```diff
@@ -45,6 +45,13 @@ fi
 
 WARDEN_PHPMYADMIN_ENABLE="${WARDEN_PHPMYADMIN_ENABLE:-1}"
 if [[ "${WARDEN_PHPMYADMIN_ENABLE}" == 1 ]]; then
+    if [[ -d "${WARDEN_HOME_DIR}/etc/phpmyadmin/config.user.inc.php" ]]; then
+        rm -rf ${WARDEN_HOME_DIR}/etc/phpmyadmin/config.user.inc.php
+    fi
```
ihor-sviziev, line 50: I don't think you can remove this directory when the container is running.

Could you confirm that?

Since changes with phpmyadmin isn't added to release yet, I suggest not covering this case at all (remove these lines), but provide a manual instruction for those who affected.

ihor-sviziev: Just checked - on Mac it worked fine. If it also works fine for other OS, I don't mind keeping it as is

monteshot: This case works even when the whole .warden directory is deleted. 
I tested it when SVC containers ran: I removed the ".warden" directory and started SVCs again.

Running the command after the incorrect type of entity is created:
image

I propose to leave all the lines in the PR.

Issue #846, opened 2025-03-01
[Command] Vite HMR not working within Laravel container


bap14:
### Version of Warden

0.15.0

### Operating System and Installation Method

macOS 15.3.1 via Brew

### Describe the Bug

Since Laravel 9 Vite has been used as the bundler; however, running the bundler in dev mode within the Warden container. Currently you have to either customize the laravel project's warden config or run just run each build after editing code.

### To Reproduce

1. Initialize a new Laravel Warden environment
2. Create a new Laravel React Starter Kit project
3. Run `npm run dev` from within warden shell
4. View login page

### Expected Behavior

Login page renders and re-renders as you modify your source code. A simple test would be to change the layout of the auth page.

### Additional context

Actual result is a blank page because requests to Vite for dynamically bundled resources are blocked or otherwise fail.

bap14:
Main PR: https://github.com/wardenenv/warden/pull/847
Documentation PR: https://github.com/wardenenv/docs/pull/29

PR #847, opened 2025-03-01
[Feature] Add support to run vite in warden container and still have HMR

bap14:
**Check List**
- [x] Is there an existing feature request / issue? #846
- [x] Matching PR in the documentation repo ( https://github.com/wardenenv/docs/pull/29 )
- [x] Entry in CHANGELOG.md

**Is your feature request related to a problem? Please describe.**  
Running `warden shell -c 'npm run dev'` in a Laravel react project using vite will start to fail rendering pages. This is because there is no routing from the browser to the Vite listener in the container.

**Describe the solution you've submitted**  
Add a new route in Traefik specifically for Vite to route requests to the vite dev server in the php-fpm container

**Describe alternatives you've considered**  
Adjusting the warden service to allow specifying additional ports you want to bind to, and then configuring the routing at the environment level so requests to port 5173 go to the vite HMR.

**Additional context**


navarr:
At a glance this looks good. I trust your testing for merging purposes and I can include it in the next release. An extra thank you for including a documentation PR!

My only feedback would be to ask if it would make sense for these projects to have vite Auto-start in its own container based on a flag. Though this is definitely a good addition

bap14:
@navarr  I haven't tested just these changes. I ported changes I made to a local install into a PR, but this is only what was necessary to get it working. I'll give it a test and report back.

As for your other feedback about this being a separate image, I see why it would make sense. A small node container running just the vite stuff would be a solution. The problem I see is the dev experience. Once you get the project files, you now have to do something like `warden env exec -it node-vite npm run dev`; or, we would need to add a new command to warden so we can do something like `warden env vite npm run dev`. Either way, it's not great.

My thought process was since we already bundle Node into our PHP image, why not leverage it and keep it simple? `warden shell -c 'npm run dev'`. To your point of separation of concerns, if we "force" the laravel package to use a separate node server for bundling, should we do the same for Magento's bundler? Does it need access to PHP or would it work so long as the full codebase is available?

I'm happy to explore that option, but this was a quick way to get this working.

navarr:
Let's move forward with this - just wanted to ask the question! 

There's a lot more Warden can do for dev experience. Thank you for the contribution!

bap14:
@navarr  I can confirm this is working: https://www.loom.com/share/b9a9b5c794834c0bbd7693b30e93d31d?sid=425998cf-8c7e-466d-9a4b-113414c9f3ad

Issue #860, opened 2025-04-09
Add Support for Valkey


navarr:
Magento 2.4.8 supports Valkey for cache.  Valkey support should be added to Warden.

### Tasks
- [x] Create Valkey images in Warden image repository
- [ ] wardenenv/warden#861 

### Discussed in https://github.com/orgs/wardenenv/discussions/859

<div type='discussions-op-text'>

<sup>Originally posted by **lbajsarowicz** April  9, 2025</sup>
Magento 2.4.8 requires valkey/valkey:8.0
Let's introduce that to the list of supported images.</div>

ptrollinspt:
https://experienceleague.adobe.com/en/docs/commerce-operations/installation-guide/system-requirements

Redis - Valkey 8

PR #861, opened 2025-04-09
Create Valkey env variables, docker partials, and cli commands

navarr:
Ticket: #860 

To enable Valkey, the new environment variables have been added:
- WARDEN_VALKEY_ENABLE
- VALKEY_VERSION

**Check List**
- [x] wardenenv/docs#36
- [x] Entry in CHANGELOG.md
- [x] Add `valkey` command that works like the `redis` command but for Valkey (or should the `redis` command intelligently work?)

**Is your feature request related to a problem? Please describe.**  
Magento 2.4.8 moves Magento from Redis to the Valkey fork, with explicit support for Valkey and a lack of Redis on the requirements table.  This PR adds support for VALKEY to Warden, while maintaining support for Redis. 

**Describe the solution you've submitted**  
I considered whether to use the approach we used for MySQL (with the distribution type and distribution version), but have decided to take the approach we did with ElasticSearch v OpenSearch; where both Redis and Valkey can independently be enabled.

If a Warden `.env` file contains `WARDEN_VALKEY=1`, Valkey will be enabled (and Valkey is not enabled by default).

If a Warden `.env` file contains `VALKEY_VERSION` it will use that as the tag for the Valkey image from the Warden repository.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 
 **Big Changes:**
 * PHP SPX support ([#820](https://github.com/wardenenv/warden/pull/820) by @SamJUK)
-* phpMyAdmin added to Warden core services ([#801](https://github.com/wardenenv/warden/pull/801) by @monteshot)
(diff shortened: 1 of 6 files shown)


navarr:
Built and tested 👍

PR #880, opened 2025-07-31
Strip MariaDB sandbox command on DB import

samsteele:
This PR strips the MariaDB sandbox command that is inserted into the top of database dumps taken from recent versions.

`/*!999999\- enable the sandbox mode */`

Ref: https://mariadb.org/mariadb-dump-file-compatibility-change/

This command can cause an error on `warden db import` when exporting/importing between different MariaDB versions.

ERROR at line 1: Unknown command '\-'.


samsteele:
Closed, don't like the idea of indiscriminately removing a security feature

